PR #11, opened 2015-02-03
Update update.info

nomadturk:
Replaced source with the one with 0.0.0.0 as default IPs


StevenBlack:
Thanks @nomadturk.  Sorry for the long delay on this!

PR #23, opened 2015-09-10
Update readme.md

mpsOxygen:
Added two lines for linux. Init scripts no longer work in some distros.


StevenBlack:
Thanks @mpsOxygen !

PR #30, opened 2015-10-23
Explicitly lowercase hostname to prevent mixed-case duplicates.

qqo:
While trying to use the list as a base for bind9 config files (and zone names are case-sensitive in it), i've found few duplicates slipped in, e.g.:

0.0.0.0 adsatt.abcnews.starwave.com 
0.0.0.0 Adsatt.ABCNews.starwave.com

The proposed simple patch fixes that.

### DIFF
--- a/updateHostsFile.py
+++ b/updateHostsFile.py
@@ -188,6 +188,7 @@ def normalizeRule(rule):
 	result = re.search(r'^[ \t]*(\d+\.\d+\.\d+\.\d+)\s+([\w\.-]+)(.*)',rule)
 	if result:
 		target, hostname, suffix = result.groups()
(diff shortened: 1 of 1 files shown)


qqo:
You're welcome!
However, please be aware that last commit (10c4449) fixes only the most obvious part of the bug, the hidden one might still exists, the data sources are just filtered enough not to trigger it.I just tried not to interfere much with code since i don't quite get what 'suffix' is for and why its needed in the first place.

But that last patch indeed may break things for some users, as it will strip IDN TLD, since \w in ([\w\.-]+)(.*) won't match it unless UNICODE is specified:

127.0.0.1 domain.日本語
will become
0.0.0.0 domain. #日本語

But i guess thats outside of the scope for now ;-)

PR #31, opened 2015-10-26
Zaursmod

qutorial:
UPD: I have merged your changes in another branch, please, disregard this pull request and see another one.

Hello and thanks for accepting the previous pull request. 

I would like to introduce here a relatively large number of modifications I did to your project.
I am pretty confident, that many people might find these mods wanted, so I would like to take my time, and a bit of your attention, please, to have a look on the changes I propose here. Below I describe them briefly.

1) Everything is converted to Python 3 and tested to work

2) A new feature is added: once someone makes a preamble.txt file, it will be appended in the beginning of a new hosts file. This might be necessary, if the hosts file contains some custom records before the "bad" hosts.

3) I add a number of hosts file providers, so the hosts file grows a lot. 300 000 hosts are blocked now.  + I added also some hosts with ads there, which I personally find bad.

4) Suffixes are not written any more in the target file. E.g. "127.0.0.1 badsite.com:443"  will be "127.0.0.1 badsite.com", this affects only two rules in the whole set. It is different from writing them into a comment.

5) The default option to exclude certain websites is set to N instead of y.

I am open for discussion!
Regards!


StevenBlack:
Hi @qutorial, I have a question.

I'm not really a Python person, so I don't know the answer.

How to avoid problems with people who have Python 2 installed?

Should I have both Python 2 and Python 3 versions in parallel for awhile?  

I'm curious how other Python-based projects are handling this.

Jamesits:
You can write a python script which is compatible to both python 2 and 3.

qutorial:
@StevenBlack  Hello, Steven. @Jamesits  is right, it is possible to make code run on both versions.  We could do this. However, this page https://wiki.python.org/moin/Python2orPython3 recommends to use Python3 which is the current version of Python, unless there is a need to use Python2 (e.g. lack of a library in Python3).

There are systems which have only Python 2 installed by default (Macs). They can install Python 3.

It is possible to have both Python3 and Python2 installed on one machine.

qutorial:
@StevenBlack I supported both Pythons now. Should work on 2 and on 3 as well. Tested on my Ubuntu with 2.7 and 3.4.

PR #32, opened 2015-10-26
Updated Pull Request

qutorial:
Update 1: I have merged your changes and changes from qqo's pull request into this branch, so this pull request is just ready to be merged.

Update 2: I have added backward compatibility to python2

Update 3: The script invokes whatever python is there on the machine

Original message:

Hello and thanks for accepting my previous pull request. 

I would like to introduce here a relatively large number of modifications I did to your project.
I am pretty confident, that many people might find these mods wanted, so I would like to take my time, and a bit of your attention, please, to have a look on the changes I propose here. Below I describe them briefly.

1) Everything is converted to Python 3 and tested to work. This change required some syntactic corrections, and almost none semantic ones.

2) Unused functionality deleted: Unzipping was removed, as it is not used actually by the hosts providers described in data. I can add it back, if needed. Re-writing it in Python 3 shouldn't take a lot of effort.

3) A new feature is added: once someone makes a preamble.txt file, it will be appended in the beginning of a new hosts file. This might be necessary, if the hosts file contains some custom records before the "bad" hosts: local LAN resources, fixed IPs for some important sites.

4) I added a number of hosts file providers. They are taken mainly from Unbeloved Hosts Android project.
As the result the hosts file grows a lot. 300 000 hosts are blocked now.  I tested it on my Ubuntu machine, and it works just fine.
+ I added also some hosts with ads there, which I personally find bad.

5) Suffixes in the rules are not written any more in the target file, even as comments. 
E.g. "127.0.0.1 badsite.com:443"  will be "127.0.0.1 badsite.com", this affects only two rules in the whole set in fact. I feel it might be safer to do so. Logging is improved to indicate such cases of suffix truncation.

6) The default option to exclude certain websites is set to N instead of y. Cause I do not watch hulu :)

I know that the modifications are quite big, and I am open for discussion! Please do not hesitate to contact me, if something is not clear in what I did.
Best regards!


StevenBlack:
Hi @qutorial thank you for this pull request.

I have several problems with this that I hope you can fix.

Can I ask you to please do the following things?

* Please separate this pull request into several distinct pull requests, each addressing a single concern.  In particular

  * The Python 2 and Python 3 PR should be distinct and separate.
  * The addition of 360,000 hosts should be a distinct pull request.
  * Everything involving `preamble.txt` file should be a separate discussion.  I have no interest in putting those personal hosts into this repo.
  * I have no need for files named `.orig`.  You should be using git branches for that.
  * None of the pull requests should be for branches named `master`.  Please name each branch appropriately for the feature.  Testing and working with alternate branches named `master` is a monumental pain.

Thanks for taking the time to put this into bite-sized pieces for consideration.

PR #33, opened 2015-10-26
Migration to python3

qutorial:
Ok, here is the first portion of work in two commits.

The first commit shows the modifications needed to port to the python3. They are there to just show how the porting was done. 

The first commit deletes support for zip files as well, as this code is dead, and porting it makes no sense.

Otherwise, the behavior is not changed (presumably).

The second commit adds backward-compatibility to python 2. This is a tested commit which works on both pythons.

Would this pull request be accepted, I could continue to implementing the other features on top of it.


StevenBlack:
Works great in Python 2.7 :smile:

PR #34, opened 2015-10-30
Update to include preamble handling

qutorial:
This update includes a feature to formulate a preamble. 

A file called myhosts is added. It is empty and ignored by git to stay so.

The user could put his custom hosts file contents inside this file for them to appear again and again in the generated hosts file without overwriting them.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+myhosts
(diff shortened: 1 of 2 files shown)


StevenBlack:
Perfect!

PR #36, opened 2015-11-04
Delete .DS_Store from Data

FrancoGag:
If the Data folder is opened in OSX it creates a .DS_Store file and the python script tries to use it a source and returns errors, deleting it solves this issue.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 myhosts
+whitelist
--- a/updateHostsFile.py
+++ b/updateHostsFile.py
@@ -16,6 +16,7 @@ import string
 import subprocess
 import sys
 import tempfile
+import glob
 # zip files are not used actually, support deleted
 # StringIO is not needed in Python 3
 # Python 3 works differently with urlopen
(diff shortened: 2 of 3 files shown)


StevenBlack:
Hi @FrancoGag and thank you for this Pull Request.

I wonder if it's better to just ignore the .DS_Store file, rather than delete it.  Maybe, as a general rule, just ignore `.` files.

FrancoGag:
That is an even better solution, I'll try to code it later

FrancoGag:
Now it no longer deletes .DS_Store and it ignores all hidden files.

I wanted to make a new pull request that included an external whitelist, I'm new with github so I didn't knew it will appear here. Sorry for that.

StevenBlack:
Can someone please test this in Python 3?

Good work @FrancoGag !

FrancoGag:
I already tried it with python 3.5 and it worked fine.
Maybe we should rename removeDups() to removeDupsAndExlc() or something like that.

StevenBlack:
:shipit:

PR #43, opened 2015-11-15
Check if 'hosts' file exists

FrancoGag:
We need to check that the file exists before trying to update it, and if it doesn't we try to create an empty 'hosts' file.

### DIFF
--- a/updateHostsFile.py
+++ b/updateHostsFile.py
@@ -102,6 +102,15 @@ def main():
 
 # Prompt the User
 def promptForUpdate():
+	
+	# Create hosts file if it doesn't exists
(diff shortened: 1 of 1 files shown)


FrancoGag:
I found about this error while trying to implement the whitelist (renaming the file to compare it and see if it was working as expected).
A more realistic case would be a user that has to manually move the host file. If they move the hosts file instead of copying it, they'll get an error.

I'll delete the `<sys.exit()>` cause I don't know if it is needed. I added it to avoid errors after displaying: "ERROR: No 'hosts' file in the folder, try creating one manually"

StevenBlack:
Awesome!  Thank You!

PR #44, opened 2015-11-15
OSX DNS flush to cover El Capitan

mariusv:
Updated the readme to cover OSX El Capitan for flushing DNS


StevenBlack:
Awesome, thanks @mariusv!

Issue #41, opened 2015-11-14
Inconsistent Merging Process


hd074:
Hello,
First I have to thank you and all constributors for your great script. So.. Thank you very much!

----
I found out that the merging process of the remote lists (adaway, mvps, etc) often [always?] results in a "different" new hosts file. I stick to your given lists. No Exclusions, no personal list added.

Behavior is the following:
1. Run Script. Update, no exlusion, dont copy
2. ```wc -l hosts ``` 30385 (Lines in the new hosts file)
3. ```cat hosts | grep "^0" | wc -l``` 28090 (That are the "pure" filter rules)

then

1. Run Script again. NO update, no exlusion, dont copy
2. ```wc -l hosts ``` 30417
3. ```cat hosts | grep "^0" | wc -l``` 28120 

After each test I checked the local remote hosts files:
```wc -l data/*/hosts``` both times the results were the same. What means, that each merge was based on the same base files. 

I did that a couple of times with and without update and even if the local remote hosts files are the same (same linecount) I get a different output (new host file).

The output "new shiny host file with X unique entries" IS always the same.

----

I ran a couple of tests on Windows 7 with cygwin (Python 2.7.10-1) and a Raspberry Pi (Raspbian Wheezy, Python 2.7.3). I always get different new hostfiles on both Devices. Updating the remote Hostfiles on both Devices and merging always leads to different newly merged hostfiles even if ```wc -l data/*/hosts``` gives the same output on both devices. Seems to be independent from the update process and device because the test in the upper section leads to the same result (different hosts file).

----

Since i'm not (yet) familiar with python a can't fix the problem myself and give feedback at this point.

hd074:
Okay i found out something more. 

After each merge the count of ```cat hosts | grep "^0" | wc -l``` is inceased by 15.

´´´sort hosts | uniq -c | sort -nr | head -30´´´ gives you 15 entries that occur ~40 times. If you merge again this number increases by 1 per entry.

so... somehow you always keep that old entries after each merge and add them again.

since i pulled the latest version from your repo and did a hard reset the should be 15*39 entries that can be permanently removed. that also means, that you merged the list ~40 times since that bug occured.

hd074:
possible hotfix:

```echo "127.0.0.1 localhost" > newhosts; sort hosts | uniq -u | grep "^0" >> newhosts```

This will delete duplicates, remove artefacts and store it in a new file "newhosts" but since it will remove all the comments this can not be considered a real solution.

[...]

PR #45, opened 2015-11-15
hotfix for duplicates and artefacts

hd074:
Issue #41

Merging with an already existing hosts file leads to duplicates and artefacts.

Now: Generate hosts file from the scratch
This is just a little fix that will prevent the occurence of that behavior.
This should not be considered a permanent solution.


StevenBlack:
Thank you @hd074!

PR #46, opened 2015-11-21
Don't print '::1 localhost'

FrancoGag:


FrancoGag:
I'll try to be as detailed as possible.
The 'hosts' file generated should be the exact same file that is generated without this commit.
The problem is that `==>::1 localhost<==` is printed in the Terminal / Command Prompt output.
I tested it and the generated hosts file has the same md5 sum with or without the commit, as it should be.

Everything that is not handled by the function `removeDupsAndExcl()` is printed in the command prompt. For example comments or empty lines ( you can try it by deleting the part that checks for comments and they'll show up in the Terminal window and not in the 'hosts' file).
Three of the sources have hosts files that include `::1 localhost` in them and it wasn't handled by  `removeDupsAndExcl()`  

The output before was:
>Do you want to update all data sources? [Y/n] n
OK, we'll stick with what we've  got locally.
Do you want to exclude any domains?
For example, hulu.com video streaming must be able to access its tracking and ad servers in order to play video. [Y/n] n
OK, we'll only exclude domains in the whitelist.
**==>::1 localhost<==**
**==>::1 localhost<==**
**==>::1 localhost<==**
Success! Your new hosts file has been prepared.
It contains 26,932 unique entries.
Do you want to replace your existing hosts file with the newly generated file? [Y/n] n

And it should be:
>Do you want to update all data sources? [Y/n] n
OK, we'll stick with what we've  got locally.
Do you want to exclude any domains?
For example, hulu.com video streaming must be able to access its tracking and ad servers in order to play video. [Y/n] n
OK, we'll only exclude domains in the whitelist.
Success! Your new hosts file has been prepared.
It contains 26,932 unique entries.
Do you want to replace your existing hosts file with the newly generated file? [Y/n] n

I hope I was clear, also I don't know if you want to include `::1 localhost` in the generated 'hosts' I don't know if there is any benefit (currently it's not included in the generated 'hosts' and it works fine as is).

StevenBlack:
Makes perfect sense.  Thanks!

PR #53, opened 2015-12-29
Recommend to remove Google Analytics record

kraaden:
I would recommend to remove analytics.google.com from the list because it makes impossible to use google analytics. I didn't change your python script or something else, but just removed the record from the host file.

### DIFF
--- a/hosts
+++ b/hosts
@@ -27990,7 +27990,6 @@
 0.0.0.0 amazon-adsystem.com
 0.0.0.0 amung.us
 0.0.0.0 anahtars.com
-0.0.0.0 analytics.google.com
 0.0.0.0 analytics.yahoo.com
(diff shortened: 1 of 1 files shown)


StevenBlack:
Hi @kraaden,

The hosts file is generative so removing the record has no effect because, upon next regeneration, it will appear again.

Take note that there is a file named [whitelist](https://github.com/StevenBlack/hosts/blob/master/whitelist) in the repo.

So you can clone the repo,  place this record (and any others) in the whitelist and it won't appear in the generated hosts file when you run `python updatehostsfile.py`.

One of the roles of this hosts file is to eliminate tracking, by Google and others.

I think the whitelist gives you options to play it as you see fit.

Thanks for the Pull Request, though.  Closing.

PR #55, opened 2016-01-03
Jonadabe patch 4

Jonadabe:
Adding hosts-file.net


StevenBlack:
The thing with hosts-file.net is, that's a massive hosts file.  It does not appear to be actively curated as such a large file would require.

It seems to be a going concern, but compared to other sources, its curation activity doesn't seem commensurate with volume.

What are your thoughts on this?

FadeMind:
:+1:  for add this

FadeMind:
@StevenBlack 
Additional hosts file are too on http://hostsfile.org/hosts.html too. (but in ZIP format)

Jonadabe:
@FadeMind 
The domain is http://hosts-file.net/ 
;-)

hd074:
I added this source in my fork some time ago.

Concerning the curation of the file I'm don't really see a problem here.
Problems that may occur on a not well maintained large file may be:

1. false positives
2. abandonment of the file
3. entries that are outdated (domain name not existing anymore)

but

1. We have a whitelist if a site breaks
2. Since this file contains a massive amount of entries there is (right now) no alternative for these entries
3. In some Issue of this repository that point was already discussed and said to be immaterial

Another important fact in connection with the mentioned points is that the much bigger file size does not seem to have too big influence on performance.

StevenBlack:
OK @hd074, you convinced me.  Thanks Thomasz @FadeMind!

StevenBlack:
I've just removed hosts-file.net as a data source in this repo.  

I've updated the readme to point to that file for those who want to include it in their own forks and clones.

PR #69, opened 2016-02-12
Made a list a list..

lipis:


lipis:
may I ask you why do you have/need the `readme_template.md` in a first place?

PR #70, opened 2016-02-12
add service restart command for Arch Linux

likyng:
Added the command for restarting the service when using Wicd, a popular alternative to NetworkManager.


StevenBlack:
Hi @likyng this is a good pull request EXCEPT the readme is generative — values get poked-in at generate-time.

Therefore can I ask you to please submit this same change to the [readme_template.md](https://github.com/likyng/hosts/blob/master/readme_template.md)?

I could easily do it, but then you won't get the credit in git history :-)

Thanks!

PR #76, opened 2016-02-14
Removed template

lipis:
I'm sure it could be done better.. but I think it's cleaner than having a full copy of readme.md just for one number..


lipis:
and for these updates.. #66 #69  no need to update two files from now on..

StevenBlack:
Thanks for the suggestion and code, @lipis.  

I won't be making this change at this time.

lipis:
oki :)

Issue #64, opened 2016-02-03
Request: Backups


estin1985:
The updateHostsFile.py should create a backup of the users original hosts file.  This is just good practice, and could prevent losing a custom hosts file built over time.

StevenBlack:
This is an excellent idea.

gigaslayer131890:
I would guess this issue can be closed now?

PR #80, opened 2016-02-14
Backup of old hosts files

gigaslayer131890:
Modified the removeOldHostsFile() function to keep backups of the old hosts file.
This is a proposed enhancement in line with the following issue
https://github.com/StevenBlack/hosts/issues/64

### DIFF
--- a/updateHostsFile.py
+++ b/updateHostsFile.py
@@ -15,10 +15,12 @@ from __future__ import absolute_import, division, print_function, unicode_litera
 import os
 import platform
 import re
+import shutil
 import string
 import subprocess
 import sys
 import tempfile
+import time
 import glob
 import argparse
 # zip files are not used actually, support deleted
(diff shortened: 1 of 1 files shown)


StevenBlack:
Hey thanks @gigaslayer131890!

StevenBlack:
@gigaslayer131890 since this brings in `shutil` what's your take on changing [this line](https://github.com/StevenBlack/hosts/blob/master/updateHostsFile.py#L338) in method `moveHostsFileIntoPlace`?

My guess: nah, leave it.

gigaslayer131890:
Thanks for accepting. This was my first open source contribution, so I am glad it went so well.
I would also agree that you can leave that line be. It works, after all. Also, after a quick search, I found out that shutil is slower than subprocess, so you already have the code optimized.

StevenBlack:
That's great @gigaslayer131890.  They say you always remember the first time :-)

PR #85, opened 2016-02-16
added ptbr info

muthdra:


StevenBlack:
Hey @muthdra can you please submit this with only the files you've changed?  In this case I'm thinking just one change to `readme_template.md` pointing to your translation.

PR #99, opened 2016-03-09
added tyzbit hosts

tyzbit:
Includes Windows 10 tracking sites not caught by previous entries.

source: https://github.com/10se1ucgo/DisableWinTracking


data/tyzbit/update.info
```diff
@@ -0,0 +1 @@
+https://raw.githubusercontent.com/StevenBlack/hosts/master/data/StevenBlack/hosts
```
StevenBlack: Hi @tyzbit  the update.info file should point to the source of this list so future updates can be incorporated.

Thus URL above ^^^^ isn't correct.   Right?

StevenBlack:
I like this!   But the `update.info` file is incorrect...

PR #103, opened 2016-03-13
add extension lists from https://github.com/pi-hole/pi-hole

nodiscc:


StevenBlack:
Thanks you @nodiscc for this pull request.

I have to reject it, however, because it's too broad.

The goal of this repo isn't to merge all the hosts we can find.  Rather it's to strike a balance of a few high-quality, actively curated lists to try and minimize Type I and Type II errors.

Can I ask you, instead, to

* Submit each source as a separate pull request, and
* investigate each source you submit for signs of active, ongoing curation by its maintainers.

I'm sure there are good sources here.  Let's pick those.  Which looks to be the best?  How would you rank the top-3, say?

nodiscc:
> The goal of this repo isn't to merge all the hosts we can find. Rather it's to strike a balance of a few high-quality, actively curated lists to try and minimize Type I and Type II errors.

Hi, I understand that these list have not been fully tested and therefore are not good candidates as default lists. I have put them as extension lists so that users can decide for themselves if they want to enable them.

> I'm sure there are good sources here. Let's pick those. Which looks to be the best? How would you rank the top-3, say?

I will investigate whether the lists are maintained, and if they are of a good quality, but maintainers at https://github.com/pi-hole/pi-hole could provide more information than I can, since they are more familiar with the lists (they have disabled some of them, explicitly stating that they are blocking legitimate domains, which I reported in the readme).

Ping @dschaper @jacobsalmela @Mcat12 @PromoFaux

The source for these hosts files was https://github.com/pi-hole/pi-hole/blob/master/adlists.default. Pi-hole also uses your hosts file as you can see in their default list.

I can either resubmit the lists, each in it's own Pull Request, but testing will take time. Or we could just reference them in the `extensions/readme.md` file, which is fine. Users would have a reference of available lists, and can add them manually as extensions.

What do you think?

Mcat12:
The best way to add optional lists is to add them to adlists.default commented out with an explanation.

PromoFaux:
We started using this list as we used to pull them all in individually, made sense to use a ready-processed one :)

As for the others, some of them are quite aggressive, but I think the ones that we gave right now are a good balance. Some of them still contain legitimate domains (e.g for some reason they block anything related to Microsoft, even Microsoft.com....) 

It's a constantly evolving list of sources that is subject to change at any time, any we've found issue with are commented out by default.

PR #105, opened 2016-03-18
Added dnsmasq and better messaging

itachi46:
* Added restarting dnsmasq as a way to flush DNS cache (used by centos/rhel/fedora camp)
* Added success messages to services being restarted
* Added an error message if none of the conditions were met instead of
silently failing

I wasn't sure if we could change some of the `if os.path.isfile` statements to `elifs`? is there a case where multiple of these services would be enabled?

I did not really touch OSX side of the things as I am not sure what the
commands do and I do not have access to an OSX box.

### DIFF
--- a/updateHostsFile.py
+++ b/updateHostsFile.py
@@ -407,42 +407,42 @@ def updateReadme(numberOfRules):
 
 def moveHostsFileIntoPlace(finalFile):
     if os.name == 'posix':
-        dnsFlushOccured = False
+        dnsCacheFound = False
         print ("Moving the file requires administrative privileges. " +
                "You might need to enter your password.")
         if subprocess.call(["/usr/bin/sudo", "cp", os.path.abspath(finalFile.name), "/etc/hosts"]):
             printFailure("Moving the file failed.")
         print ("Flushing the DNS Cache to utilize new hosts file...")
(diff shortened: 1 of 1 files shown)


StevenBlack:
Thanks for this, @itachi46.  I'll be considering this shortly — heavy day ahead.

In the meantime, can I ask you to evaluate the semantics of `dnsFlushOccured = True` ?  

At the junctures where `dnsFlushOccured = True` is used, we can only say that a folder is present.  The flush, *per se*, hasn't happened yet, and it could fail.

Maybe the variable is better named `dnsCacheFound`?

itachi46:
No worries @StevenBlack, I completely agree with the variable name change I'll make a commit to change the variable over.

qutorial:
I am not sure you need to restart dnsmasq to update the DNS cache. It works without any restart of dnsmasq on my machine on the hosts file change. Do you write the dnsmasq conf files now?

itachi46:
@qutorial that doesn't seem to be the case for me the reason I made this fix was the script said it was flushing my dns but I was still able to nslookup some of the hosts that were blocked from my clients.

StevenBlack:
Thanks @itachi46 !

Issue #109, opened 2016-03-18
127.0.1.1     <hostname> is missing in generated hosts file


matkoniecz:
Default hosts file on Ubuntu (and probably also other Linux variants) has by default line

> 127.0.1.1     <hostname>

in hosts file. In addition one this file is missing system may misbehave (starting from problems with sudo).

It would be relatively easy to add this line on Linux but is it also necessary for other OS? Maybe also Macs?

mad4coinz:
Is this something that should be updated in the documentation? A note for ubuntu and other users to add this line to their "myhosts" file?
I was a bit weary of having my whole hosts file generated by this script but luckily the first few lines on my mac matched what was generated by the script. 
Otherwise we may need to add either :

a) Intelligence to identify different configurations and create the correct default hosts file based on this
or
b) add a flag to generate a "myhosts" file or special additions file the first time the script is run based on the existing hosts file and do away with the section that adds the default stuff.

FadeMind:
@matkoniecz @StevenBlack 
Reference: https://www.debian.org/doc/manuals/debian-reference/ch05.en.html#_the_hostname_resolution

matkoniecz:
@mad4coinz This script currently attempts to include definitions appearing on various OSes - https://github.com/StevenBlack/hosts/blob/master/updateHostsFile.py#L379

@FadeMind 

> For a system with a permanent IP address, that permanent IP address should be used here instead of 127.0.1.1.

may be problematic, though given 

> The Debian Installer creates this entry for a system without a permanent IP address as a workaround for some software (e.g., GNOME) as documented in the bug #719621.

it seems that `127.0.1.1 <hostname>` is one that is generated by default.

StevenBlack:
Does everyone feel the merge of #110 address this satisfactorily?

matkoniecz:
> Does everyone feel the merge of #110 address this satisfactorily?

I may confirm that adding `127.0.1.1 <hostname>` is not required for win 7. Its default hosts file has

>  # localhost name resolution is handled within DNS itself.
>  #	127.0.0.1       localhost
>  #	::1             localhost

The same situation applies to win 8, win 10 and probably other win OSes - see https://support.microsoft.com/en-us/kb/972034 and http://windowsreport.com/reset-host-file-default-windows-10/

StevenBlack:
Closing this now.

PR #110, opened 2016-03-18
include 127.0.1.1 <hostname> line

matkoniecz:
partial, maybe full fix for #109 

Tested on Python 3.4.3 and Python 2.7.6 using Lubuntu 14.04.


StevenBlack:
Nice!  Thanks!

PR #111, opened 2016-03-18
replace 'prepend' by more true 'added near top'

### LINKED ISSUE #108: preamble is not prepended
According to the readme 

> If you have custom hosts records, place them in file myhosts. The contents of this file are prepended to the unified hosts file during the update process.

but it is added after some other entries (see https://github.com/StevenBlack/hosts/blob/master/updateHostsFile.py#L366).

For example 
>  # This hosts file is a merged collection of hosts from reputable sources,
>  # with a dash of crowd sourcing via Github

is before any content from myhosts.

matkoniecz:
fixes #108

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -89,7 +89,7 @@ Add one or more *optional* extensions, which originate from subfolders of the `e
 
 ## How do I incorporate my own hosts?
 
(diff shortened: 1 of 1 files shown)


StevenBlack:
Thanks @matkoniecz !

PR #112, opened 2016-03-18
make -n working not only in combination with -a

### LINKED ISSUE #107: script run with --noupdate should not ask "Do you want to update all data sources?"
> python updateHostsFile.py -n
> ←[94mDo you want to update all data sources?←[0m [Y/n]

the culprit is

> response = "yes" if auto else query_yes_no("Do you want to update all data sources?")
>     if response == "yes" and update:

query_yes_no should not be asked for update=False set by `update = not args.noUpdate`

matkoniecz:
fixes #107


StevenBlack:
Thanks again @matkoniecz !

PR #123, opened 2016-03-31
Fixed file not found error due to relative path

itachi46:
When updateHostFile.py is not run from the project directory we get a file not found error trying to find readmeData.json


itachi46:
gonna be honest here and say my goal was to just fix the error (I run this as a cron) I noticed other files being referenced by relative path too but did not really have a look how they are used

StevenBlack:
I like!  :+1:

PR #129, opened 2016-04-19
Add option to skip static localhost entries in the final hosts file.

indrajitr:
This is particularly handy when you want to add the generated hosts file as additional _static hosts_ file in a DNS server without affecting the default `/etc/hosts`.

For example, `--addn-hosts=<file>` in [dnsmasq](http://www.thekelleys.org.uk/dnsmasq/docs/dnsmasq-man.html) can be used exactly for this purpose.

### DIFF
--- a/updateHostsFile.py
+++ b/updateHostsFile.py
@@ -81,6 +81,7 @@ defaults = {
     "freshen" : True,
     "replace" : False,
     "backup" : False,
+    "skipstatichosts": False,
     "extensionspath" : os.path.join(BASEDIR_PATH, "extensions"),
     "extensions" : [],
     "outputsubfolder" : "",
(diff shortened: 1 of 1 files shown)


indrajitr:
This PR is just an idea. If this makes sense and is applied, relevant updates need to be applied in `readme.md` as well since it would be a deviation from the documented expectation (_These unified hosts files should serve all devices, regardless of OS._).

StevenBlack:
This is an interesting pull request Indrajit @indrajitr.  This seems like a good feature.

Since it's an option, and it defaults to `false`, I can't think of any objections.

@FadeMind what's your opinion?

StevenBlack:
I like this.  Merging.  Thanks @indrajitr !

PR #131, opened 2016-04-20
Switch to https for adaway.org source

indrajitr:
Since adaway.org [does support ssl](https://github.com/Free-Software-for-Android/AdAway/blob/1000992c9fcd00a169b14b23e91d227d0cf6bfb2/hosts/hosts.txt#L8). It is still on ssl v1 though.


StevenBlack:
I dig it.  Thanks @indrajitr !

PR #141, opened 2016-06-13
Handle the auto flag in promptForFlushDnsCache

cryzed:
Forgot about this. This allows you to actually use the auto-flag without being prompted to flush the DNS cache.


StevenBlack:
Good catch!  I forgot about --auto :-)

PR #168, opened 2016-09-01
Add rexchange.begun.ru used on livejournal.com

sgtpep:


StevenBlack:
Hi Danil @sgtpep,

This repo is generative, meaning it combines several Hosts sources into a a variety of consolidated hosts file with duplicates removed.

So changing the generated file doesn't actually work since the next regeneration will lose the changes.

Three options:

* Convince one of the host sources to carry the rexchange.begun.ru domain.  This is actually easy to do.

* Create a new reliable data source that we can add to our list of sources.

* Simply add rexchange.begun.ru to the `myhosts` file and use one of the Python scripts to generate your own personal file.

I appreciate the pull request, however.  Thank you.

Closing.

PR #181, opened 2016-09-22
add.2o7Net rules

FadeMind:
@StevenBlack 

1108 additional blocking rules for **207 dot net** tracking domains.

See: 
http://www.makeuseof.com/tag/that-mysterious-2o7-net-tracking-cookie-all-you-need-to-know/
and https://en.wikipedia.org/wiki/Omniture


StevenBlack:
I dig it.  Thanks!

PR #198, opened 2016-11-02
Removing cdn.tabnak.ir

drupalshift:
This completely breaks their site , they use same cdn for their news images and website stylesheets , more reasonable approach is adding this `http://cdn.tabnak.ir/files/adv/*` pattern to ad-blockers

### DIFF
--- a/hosts
+++ b/hosts
@@ -25040,7 +25040,6 @@ fe80::1%lo0 localhost
 0.0.0.0 cdn.stat.easydate.biz
 0.0.0.0 cdn.stickyadstv.com
 0.0.0.0 cdn.syn.verticalacuity.com
-0.0.0.0 cdn.tabnak.ir
 0.0.0.0 cdnt.yottos.com
 0.0.0.0 cdn.usabilitytracker.com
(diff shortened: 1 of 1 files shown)


StevenBlack:
Hi @drupalshift, thank you for these pull requests.

As you may now know, the hosts file is *generated*, so changing the final hosts file doesn't work because a new generated hosts file will come along and replace it.

If you want to add or remove domains, you have a couple of options:

1. see [Generate your own unified hosts file](https://github.com/StevenBlack/hosts#generate-your-own-unified-hosts-file) and, therein, the [How do I include my own custom domain mappings?](https://github.com/StevenBlack/hosts#generate-your-own-unified-hosts-file "How do I include my own custom domain mappings?") and [How do I prevent domains from being included?](https://github.com/StevenBlack/hosts#generate-your-own-unified-hosts-file "How do I prevent domains from being included?") sections in the docs.

2. The `cdn.tabnak.ir` domain is aggregated from[someonewhocares.org](http://someonewhocares.org).  You could contact them and explain why you feel this domain is OK.

Closing.

PR #199, opened 2016-11-08
block mywot domain against risky addons for browsers

FadeMind:
@StevenBlack 
Web Of Trust developers selling collected data: 
http://thehackernews.com/2016/11/web-of-trust-addon.html
http://www.ghacks.net/2016/11/01/browsing-history-sold/


StevenBlack:
Thanks Tomasz! @FadeMind

PR #217, opened 2016-12-16
BAT files: change PATH for lookin DLL file check

FadeMind:
@StevenBlack


StevenBlack:
Thanks Tomasz @FadeMind !

PR #223, opened 2016-12-22
Force updateHostsWindows.bat to be launched as Administrator

crazy-max:
:gift:


updateHostsWindows.bat
```diff
@@ -1,17 +1,44 @@
...
+:ADMIN
+if exist "%TEMP%\getadmin.vbs" (
+    DEL "%TEMP%\getadmin.vbs"
+)
+pushd "%CD%"
+CD /D "%~dp0"
+CD %CD%
+%COMSPEC% /c "updateHostsWindows.bat" start
+EXIT
+
```
StevenBlack: Well!  How cool is that?  @crazy-max is on fire!

FadeMind:
@crazy-max @StevenBlack 

For me (Windows 10 here) BAT file was launched as admin via UAC prompt without any data in window (only blink white space not seen here - sorry).
image 001

crazy-max:
Yes there is an issue with the args, i will create a new PR to solve this

PR #227, opened 2016-12-26
script for update Windows DNS server config

klim-iv:


ScriptTiger:
Can this PR be closed? It will need to be re-forked before it can be processed and its been here since last year so I'm not sure it was a good fit to be added anyway. Conversion scripts are nice, but it ultimately leads to a rabbit hole of unlimited formats. Hosting separate repos for dedicated support to different formats delegates out some of the work and keeps the focus here on the core domain name lists which are already tens of thousands of entries long and deserve focus unto themselves to be curated correctly.

PR #249, opened 2017-02-22
Remove breitbart.com from the fakenews extension

xorgy:
Breitbart has about the same record as _The Huffington Post_, or _CNN.com_ which are not blocked.

Alternative to #250


StevenBlack:
Thanks for the PR.

Breitbart is curated here:  https://github.com/marktron/fakenew

Closing.

PR #285, opened 2017-05-08
Add .idea and .iml to .gitignore

gfyoung:
Just a small patch to add some IDE files to the `.gitignore`.

I encountered these when attempting to make a larger change to one of the scripts.


StevenBlack:
Thanks @gfyoung!

PR #286, opened 2017-05-08
Patch promptForFlushDnsCache Bug

gfyoung:
Previously, it was checking if `query_yes_no` returned `False`, which does not happen, as the function only
returns `"yes"` or `"no"`. As a result, the check was always returning `True`, even when the user did not
want to flush the DNS cache.

### DIFF
--- a/updateHostsFile.py
+++ b/updateHostsFile.py
@@ -191,11 +191,13 @@ def promptForMoreCustomExclusions(question="Do you have more domains you want to
 
 
 def promptForFlushDnsCache():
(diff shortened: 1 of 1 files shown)


StevenBlack:
Good catch!  I always want to flush :-).

gfyoung:
That is true!  I stumbled upon this by accident.

gfyoung:
As an aside, I wonder if it might be better to just return `True` or `False` (instead of `"yes"` or `"no"`).  Less prone to bugs IMO.

PR #287, opened 2017-05-08
Update hosts file in-place on Windows

gfyoung:
Add functionality to update hosts file in-place on Windows via command prompt administrator.

### DIFF
--- a/updateHostsFile.py
+++ b/updateHostsFile.py
@@ -440,14 +440,35 @@ def updateReadmeData():
 
 
 def moveHostsFileIntoPlace(finalFile):
+    """
+    Move the newly-created hosts file into its correct location on the OS.
+
+    For UNIX systems, the hosts file is "etc/hosts." On Windows, it's
+    "C:\Windows\system32\drivers\etc\hosts."
+
+    For this move to work, you must have administrator privileges to do this.
+    On UNIX systems, this means having "sudo" access, and on Windows, it
+    means being able to run command prompt in administrator mode.
+
+    Parameters
+    ----------
+    finalFile : str
+        The name of the newly-created hosts file to move.
+    """
+
+    filename = os.path.abspath(finalFile.name)
+
     if os.name == "posix":
-        print ("Moving the file requires administrative privileges. " +
-               "You might need to enter your password.")
-        if subprocess.call(["/usr/bin/sudo", "cp", os.path.abspath(finalFile.name), "/etc/hosts"]):
+        print("Moving the file requires administrative privileges. "
+              "You might need to enter your password.")
+        if subprocess.call(["/usr/bin/sudo", "cp", filename, "/etc/hosts"]):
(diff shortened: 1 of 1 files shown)


updateHostsFile.py
```diff
@@ -438,14 +438,35 @@ def updateReadmeData():
...
+        if subprocess.call(["/usr/bin/sudo", "cp", filename, "/etc/hosts"]):
             printFailure("Moving the file failed.")
     elif os.name == "nt":
-        print("Automatically moving the hosts file in place is not yet supported.")
-        print("Please move the generated file to %SystemRoot%\system32\drivers\etc\hosts")
+        print("Moving the file requires administrative "
+              "privileges in command prompt.")
+        if subprocess.call(["cmd", "/C", "copy", filename,
+                            "C:\Windows\system32\drivers\etc\hosts"]):
+            printFailure("Moving the file failed.")
```
StevenBlack: @gfyoung I'm hesitant to merge this for several reasons:
 
* So many version of Windows in userland.
* The user's rights to `C:\Windows\system32\drivers\etc\hosts`
* The Windows access control story is a steaming sewer fire.

I just fear that this potentially creates more problems than it resolves.  Your thoughts?

gfyoung: Hmm...to be fair, you do advocate using the batch files in the `README`, which does in fact copy files directly into this directory (you can't run this command unless you're administrator).  Thus, I'm not so concerned about that user's rights (your second point).

In terms of Windows compatibility across versions, I'm not sure what the target is, but I know that it works with machines as old as Windows 7.  I figured that unless someone has a machine older than this, that won't be a problem.

As for your final point, I do agree that this is a problem.  However, if people do run into this problem, we can always tell people to move the file manually unless they can provide a nice solution in code.

In short, I don't believe that that will be the case.  Let me know what you think.

gfyoung:
@StevenBlack : In light of discussion in #291, closing this.  Will transfer any clean-up changes to that PR.

gfyoung:
@StevenBlack : I have a branch on my fork called <a href="https://github.com/gfyoung/hosts/tree/windows-experimental"> windows-experimental </a> which incorporates all of the changes I've made to make the `updateHostsFiles.py` runnable on Windows (standalone).  I'm happy to maintain this, as I've found this to work quite well (at least on my machine and for several others).

If you have any interest in incorporating this into the main code-base (probably as a separate branch off `master` so that more proficient Windows users can use it), let me know!

gfyoung:
@StevenBlack :

https://support.microsoft.com/en-us/help/13853/windows-lifecycle-fact-sheet

(in reference to my comment [here](https://github.com/StevenBlack/hosts/issues/769#issuecomment-418470914))

With Windows 7 extended support scheduled to be ended in about 15 months (and Microsoft dropping support entirely for anything earlier than that), what do you think about revisiting this?

PR #293, opened 2017-05-14
Have query_yes_no return boolean

gfyoung:
Bugs have been found where we assumed that query_yes_no returned `True` / `False`.  Not only is it easier to check booleans instead of strings, but it also looks a lot cleaner in the code.


StevenBlack:
Beauty!  Tested it, couldn't break it.  Thanks @gfyoung!  ⭐️

PR #296, opened 2017-05-15
Condense Python 2-3 compatibility code

gfyoung:
Cleans up the Python 2-3 compatibility checks and removes additional unnecessary imports.

Also removed a bunch of comments regarding cross-Python implementation, as they are quite redundant and don't actually clarify much as to what is happening.


StevenBlack:
Another nice PR @gfyoung.

Thanks!

PR #297, opened 2017-05-15
Standardize underscore names in updateHostsFile.py

gfyoung:
Some of my PR's accidentally introduced underscore naming into a file that is largely camel case.  Generally, it is good (stylistically-speaking) to have a consistent naming convention.

Looking through PEP8 conventions, it seems that underscore is preferred, so I removed a lot of camel casing in `updateHostsFile.py` as a result.

### DIFF
--- a/updateHostsFile.py
+++ b/updateHostsFile.py
@@ -37,24 +37,25 @@ else:  # Python 2
     from urllib2 import urlopen
 
 
-def getFileByUrl(url):
+def get_file_by_url(url):
     try:
         f = urlopen(url)
         return f.read().decode("UTF-8")
     except:
         print("Problem getting file: ", url)
-        # raise
 
 
-def writeData(f, data):
+def write_data(f, data):
     if PY3:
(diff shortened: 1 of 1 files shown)


StevenBlack:
I'm glad you did this, @gfyoung.

On balance I would have preferred `mixedCase` over `lower_case_with_underscores` because that's most common in node/javascript where I spend most of my time.  That said, I couldn't find a compelling proscription, one way or another, with Python.

Basically, I'm glad to standardize on this based on your initiative.

Tested, works great in all facets I tried.

gfyoung:
Thanks @StevenBlack !  I took my decision from the following link <a href="https://www.python.org/dev/peps/pep-0008/#id34">here</a> for future reference.

StevenBlack:
Yes, I saw that.  It's not authoritatively stated, one way or another.  I'm happy it's standardized on something.  Thanks for that.

Next thing would be, pick a Python linter, and add a linter config file to flag code standards violations in the `.py` files.  Not a big priority, something nice to have in the future.

Issue #298, opened 2017-05-15
Pick, config a Python code linter


StevenBlack:
Arising from PR #297: it would be nice to involve a Python linter in this repo.

gfyoung:
You can use `flake8`, which you can install with `pip`.  I'm putting up a PR to fix all linting errors in `updateHostsFile.py`.

gfyoung:
If you want to enforce these style errors, what you could consider is hooking up your repository to a CI service (e.g. Travis) to perform style checks on all PR's before merging.

StevenBlack:
Nice suggestion.

gfyoung:
@StevenBlack : I have a branch that implements style check testing on Travis.  <a href="https://travis-ci.org/gfyoung/hosts/builds/232629321">Here</a> is one where the builds fail the style checks.  <a href="https://travis-ci.org/gfyoung/hosts/builds/232629794">Here</a> is one where the builds pass the style checks.

If that looks good to you, I can put this up as a PR, though you will have to link your repository to Travis (since you own the repository).

StevenBlack:
Looks good @gfyoung.  If you submit the PR then I'll link `flake8` through Travis.

gfyoung:
@StevenBlack : PR submitted, though Travis isn't triggering.

PR #299, opened 2017-05-15
Patch all style errors in updateHostsFile.py

gfyoung:
Just ran the following commands:

~~~
pip install flake8
flake8 updateHostsFile.py
~~~

There should be no output from this command if all style checks are fixed.

xref #298


gfyoung:
@StevenBlack : If this looks good to you, I can do a similar one for `updateReadme.py` afterwards.

StevenBlack:
Yeah this is great.  Merging.

PR #300, opened 2017-05-15
Patch all style errors in updateReadme.py

gfyoung:
Just ran the following commands:

~~~
pip install flake8
flake8 updateReadme.py
~~~

There should be no output from this command if all style checks are fixed.

xref #298


StevenBlack:
Beauty.

PR #303, opened 2017-05-16
Lift restriction on Python 3.x for makeHosts*

gfyoung:
1) `updateReadme.py` to be Python 2.x compatible.  Let's make all of the Python scripts cross-Python.

2) Make all of the `makeHosts*` scripts the same.  The scripts now generate the same host files regardless of  which platform you are on.

3) Lift the Python 3.5 restriction on the Windows `makeHosts` batch file.  Given that not all users use Python 3.5, this script is unnecessarily restrictive.  This change will only increase the number of users who can use it (I tested this with Python versions as far back as Python 3.3 as well as Python 2.7).

4) Remove the documentation on generating hosts files for Windows.  I don't know why this is necessary for Windows when I can run the other batch file just fine without it.

### DIFF
--- a/makeHosts
+++ b/makeHosts
@@ -1,26 +1,26 @@
 #!/bin/sh
 
 # These create various alternate hosts files by combining and adding the gambling, porn, and social media extensions.
-python3 updateHostsFile.py -a -z    -o alternates/gambling -e gambling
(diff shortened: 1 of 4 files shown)


gfyoung:
@StevenBlack : Successfully rebased and ready to go.

StevenBlack:
Thanks @gfyoung!  Merging.

PR #306, opened 2017-05-16
Lift Python 3.5.x restriction on updateHostsWindows.bat

gfyoung:
1) Lift the Python 3.5.x restriction on `updateHostsWindows.bat`.  Per the existing documentation, `updateHostsFile.py` is cross-Python compatible.

2) Greatly reduce the checks for administrator privileges.  We should be running this file as administrator on command prompt, so we shouldn't be requesting administrator access if that isn't the case (case in point: when I run the original batch file as is on normal command prompt, I get caught in an infinite loop that attempts to request administrator access).  Rather, we should just kill the script.

### DIFF
--- a/readme_template.md
+++ b/readme_template.md
@@ -210,8 +210,8 @@ sudo dscacheutil -flushcache;sudo killall -HUP mDNSResponder
 |`makeHostsWindows.bat` BATCH file will create various alternate hosts files by combining and adding the gambling, porn, and social media extensions. You need to be connected to the Internet. This file REQUIRED installed Python 3.5.x runtime environment in Windows System. Launch this file as normal user.|
 :---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 
-|Run `updateHostsWindows.bat` BATCH file in Command Prompt with Administrator privileges in repository directory for easy update, replace hosts file and reload DNS cache in Windows System. You need to be connected to the Internet. This file REQUIRED installed Python 3.5.x runtime environment in Windows System.|
-:---------------------------------------------------------------------------------------------------------------------------------------------------------------------|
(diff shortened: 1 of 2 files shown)


StevenBlack:
Good!  Thanks @gfyoung!

gfyoung:
@FadeMind : I ran the code on several machines before submitting this.  What exactly crashed and how did it happen for you?

FadeMind:
After double click on Your BAT file I saw CMD prompt window for a 2 seconds and nothing happened.
With previous code review (and after removing Py 3.5 checking) UAC prompt works fine.

gfyoung:
The documentation explicitly says you must run this in Command Prompt with administrator privileges.

FadeMind:
BUT in previous code BAT file only call UAC shell for admin rights :P So, your commit is damaged and rollback functionality.

gfyoung:
Sure, but the external documentation should take precedence.  This has been the case for quite some time.  I have a hard time accepting that we should "accommodate" functionality that was never documented NOR recommended in the first place.

StevenBlack:
💭  ah... Windows.

FadeMind:
@StevenBlack  can I pull req. for revert this with REMOVED PY ENV check func.?

gfyoung:
I would be against reverting this commit.  It's perfectly in-line with documentation, and I don't agree that we should be allowing people to run this BAT file in ways we never even recommended it to be run in the first place.

FadeMind:
Security first, Usability Third? 😁

StevenBlack:
I just want an uncomplicated ⛵️ life.

In the good old days, we just said, *"Windows user? Copy the hosts file yourself"*.  Good times.

gfyoung:
If you try to open command prompt with administrator privileges, UAC kicks in.  Currently, the script dies immediately if not run in command prompt.  I think that's pretty secure.

PR #310, opened 2017-05-18
Add DNS cache flushing calls for Linux

### LINKED ISSUE #128: Unable to determine DNS management tool. Linux Mint
For Linux Mint you can run
`sudo /etc/init.d/dns-clean start`

to run it automaticlly.

gfyoung:
1) Add command for Linux Mint
2) Add additional command for NCSD cache (ran into this one)

Closes #128.

### DIFF
--- a/updateHostsFile.py
+++ b/updateHostsFile.py
@@ -559,14 +559,17 @@ def flush_dns_cache():
         print("Please copy and paste the command 'ipconfig /flushdns' in "
               "administrator command prompt after running this script.")
     else:
-        if os.path.isfile("/etc/rc.d/init.d/nscd"):
-            dns_cache_found = True
+        nscd_caches = ["/etc/init.d/nscd",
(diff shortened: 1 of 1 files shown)


rautamiekka:
It's not only Linux Mint, it's every unmodified Debian-based distro; by default (there's a Debian Edition) LM is a polished Ubuntu (dunno what _exactly_ the default LM is based on), which in turn is an easier/"polished" Debian Unstable.

Would seem the script'll run on RHEL-based systems (CentOS, Fedora, to name a couple) but I don't have VirtualBox handy on this computer to check.

gfyoung:
@rautamiekka : I think that's right (can't test either ATM)...I specifically said Mint because that's the distro I was testing the script on (and also the topic of the related issue).

StevenBlack:
Awesome as always, @gfyoung !

PR #338, opened 2017-06-20
Tighten regex rule to discard IP address as hostname

indrajitr:
The regex rule to normalize host data in `normalize_rule` needs to be adjusted to avoid spurious entries like so:
```
0.0.0.0 130.211.230.53
```

While at it, just went ahead and tighten the regex rule to match valid IPv4 addresses only.

### DIFF
--- a/updateHostsFile.py
+++ b/updateHostsFile.py
@@ -501,7 +501,8 @@ def normalize_rule(rule):
         The rule string with spelling and spacing reformatted.
     """
 
-    result = re.search(r'^[ \t]*(\d+\.\d+\.\d+\.\d+)\s+([\w\.-]+)(.*)', rule)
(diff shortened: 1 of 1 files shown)


StevenBlack:
Hi Indrajit – @indrajitr – thank you very much for this pull request.

A question:  I think this breaks with domains that begin with a number?

Example:  [line 83 of the mvps.org hosts file](https://github.com/StevenBlack/hosts/blob/362df27f4351dac5322cc2df2e731111428d1d5a/data/mvps.org/hosts#L83) is 0.0.0.0 222-33544_999.pub.adfirmative.com

This happens to be

indrajitr:
Hi @StevenBlack,
Interesting that it didn't work for you. The only diff I see between the two generated `hosts` files is this:
```
5c5
< # Number of unique domains: 36,559
---
> # Number of unique domains: 36,558
31595d31594
< 0.0.0.0 130.211.230.53
```

In case you want to try out the variations, this might be handy:
https://regex101.com/r/0Ar0sO/1

PR #341, opened 2017-06-22
Honor whitelist exclusion only for domain or subdomain match

indrajitr:
* Whitelisting `s.youtube.com` should not whitelist `ads.youtube.com`; but can whitelist `a.s.youtube.com`
* Whitelisting `twitter.com` should not whitelist `ttwitter.com` or `twitter.com.au`; but can whitelist `a.twitter.com`

### DIFF
--- a/updateHostsFile.py
+++ b/updateHostsFile.py
@@ -471,7 +471,7 @@ def remove_dups_and_excl(merge_file):
         # Normalize rule
         hostname, normalized_rule = normalize_rule(stripped_rule)
         for exclude in exclusions:
-            if exclude in line:
+            if re.search('[\s\.]' + re.escape(exclude) + '\s', line):
                 write_line = False
                 break
 


gfyoung:
@indrajitr : Something's not quite right with that regex IINM:

~~~
>>> import re
>>> exclude = "twitter.com"
>>>
>>> line = "a.twitter.com"
>>> re.search('[\s\.]' + re.escape(exclude) + '\s', line)
>>>
>>> line = "twitter.com.au"
>>> re.search('[\s\.]' + re.escape(exclude) + '\s', line)
>>>
~~~

gfyoung:
IMO, regex matches should be tested to confirm that they do what they are supposed to do.  Not sure how best to do that, but throwing that out there for consideration.

rautamiekka:
https://regex101.com/ is your friend. It's a realtime RegEx editor, tester, explainer, highlighter and debugger (although debugging is a separate feature you need to switch to, giving up the other features; in debug mode the system goes through your strings a character at a time, further explaining it) with support for at least 4 flavors: PHP aka PCRE, JavaScript, Python, Golang.

indrajitr:
@gfyoung: The `line` would have IP address prepended already at that point in `remove_dups_and_excl` call. Also, the lines are `\n` terminating.

So your test should look like:

```python
>>> import re
>>> exclude = "twitter.com"
>>> 
>>> re.search('[\s\.]' + re.escape(exclude) + '\s', "0.0.0.0 twitter.com\n")
<_sre.SRE_Match object; span=(7, 20), match=' twitter.com\n'>
>>> 
>>> re.search('[\s\.]' + re.escape(exclude) + '\s', "0.0.0.0 twitter.com \n")
<_sre.SRE_Match object; span=(7, 20), match=' twitter.com '>
>>> 
>>> re.search('[\s\.]' + re.escape(exclude) + '\s', "0.0.0.0 twitter.com # comment\n")
<_sre.SRE_Match object; span=(7, 20), match=' twitter.com '>
>>> 
>>> re.search('[\s\.]' + re.escape(exclude) + '\s', "0.0.0.0 a.twitter.com\n")
<_sre.SRE_Match object; span=(9, 22), match='.twitter.com\n'>
>>> 
>>> re.search('[\s\.]' + re.escape(exclude) + '\s', "0.0.0.0 ttwitter.com\n")
>>> 
>>> re.search('[\s\.]' + re.escape(exclude) + '\s', "0.0.0.0 a.ttwitter.com\n")
>>> 
>>> re.search('[\s\.]' + re.escape(exclude) + '\s', "0.0.0.0 twitter.com.au\n")
>>> 
>>> re.search('[\s\.]' + re.escape(exclude) + '\s', "0.0.0.0 twitter.com.au  #comment\n")
>>> 
```

@rautamiekka: Summoning our good friend to validate :)
https://regex101.com/r/ewDQsX/1/

gfyoung:
@indrajitr : Ah, okay.  That makes sense.  I wasn't sure what the input line looked like at this point.

indrajitr:
@StevenBlack: anything else you might want for this one?

gfyoung:
@indrajitr : Hard to test this code ATM, but hopefully one day with some more refactoring (which I started and will resume), it will :smile:

indrajitr:
@gfyoung, that would be fantastic! Indeed bit of refactoring for would be useful.

Attempted that with  `remove_dups_and_excl`, and then ran out of time for now 😞.
But would be happy to help your effort since you already started.

gfyoung:
@indrajitr : Certainly, I don't think refactoring `remove_dups_and_excl` in this PR would have been prudent.  Refactoring this code isn't so bad.  However, for every refactored function, there must be tests, and that's when things get tricky :smile:

StevenBlack:
Thanks indrajit @indrajitr , looks good!

Issue #352, opened 2017-07-08
Why do all the hosts describe total unique domains as 0?


onmyouji:

Jezza-x3:
Where? Can you show what you mean?

Jezza-x3:
Nevermind, I see. In the readme. @stevenblack

onmyouji:
And also in the hosts file itself.

gfyoung:
That's my bad.  I've been trying to refactor the code a bit, but I overlooked this.

PR #354, opened 2017-07-09
Fix the number of unique entries.

meinhimmel:
Fixes issue #352 by moving the final merge before getting the number of unique rules.


gfyoung:
@meinhimmel : Thanks for catching this!  More reason to add more tests.  I overlooked that `remove_dups_and_excl` modified this, though I cannot test this function yet. :cry:

meinhimmel:
@gfyoung Ha, no problem.

PR #355, opened 2017-07-09
Drop nose dependency on Travis

gfyoung:
Nose is not a well-supported library anymore (see <a href="https://nose.readthedocs.io/en/latest/">here</a>).

`unittest.main()` is sufficient for our purposes.  Note, we can still note that it is possible in the documentation, but we shouldn't rely on it for our testing machines.

### DIFF
--- a/ci/install_conda_env.sh
+++ b/ci/install_conda_env.sh
@@ -5,4 +5,4 @@ conda create -n hosts python=$PYTHON_VERSION || exit 1
 source activate hosts
 
 echo "Installing packages..."
-conda install mock nose flake8
+conda install mock flake8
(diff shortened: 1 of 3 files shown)


StevenBlack:
@gfyoung  can we just call it `test.sh`?

gfyoung:
Sure, why not?  We could also rename `lint_repo.sh` to `lint.sh` ?  (Now that I think about it, maybe we could save the renaming for another PR?)

StevenBlack:
Which reminds me... documentation?   :smiley: 

So how about rename, and doc, in this PR?

I like raising the bar sometimes. :runner:

gfyoung:
I'm not sure I follow you here.  Did you also want to do the renaming in this PR as well?

Also, I'm not sure I follow you regarding documentation.  I never said `nose` was a dependency in the docs.

StevenBlack:
Thanks @gfyoung !

PR #219, opened 2016-12-19
Update bash script for Mac / Unix

pythonmagic:
Update bash script for Mac / Unix

PR #365, opened 2017-07-21
Simple Unix Hosts Updater Script

ValkiryXDA:
Hey im kind new to this whole thing but i made a simple idiot proof script for Linux systems. Let me know what you thing thanks


ValkiryXDA:
I think a simple bash script is a bit more viable since not everyone installs python or uses it for that matter. Thanks for you input dude

funilrys:
As a bash basher I'm not against bash @ValkiryXDA but today the most popular distributions comes with Python pre-installed so "installing python" is not really a problem ... 

You work is not bad in between  ... :wink: 

Steven @StevenBlack will have the last word so keep up the good work :+1:

HostsFileUpdaterScript
```diff
@@ -0,0 +1,47 @@
...
+echo "Retrieving $HFILE from $HFSERVER"
+echo ""
+wget -O /tmp/$HFILE $HFSERVER/$HFILE
+# Uncomment the line below to allow .zip file extraction of hosts files.  
+#unzip -p /tmp/$HFILE | dos2unix > /tmp/hosts
+if [ 'grep -c "banner" /tmp/hosts' ];then 
+    echo "Downloaded and unpacked $HFILE OK"
+    echo "Appending host list to original content"  # which was probably there for a reason, like to make sure localhost worked, and possibly even more stuff if part of a corporate LAN
+    #cp -f -u /tmp/hosts /etc/hosts
+    cat $ORIGFILE  >/etc/hosts
```
funilrys: This is an issue ... I didn't tested it but `>` will replace the file with the current read line so at the end you only have the last line of your `${ORIGFILE}`

HostsFileUpdaterScript
```diff
@@ -0,0 +1,47 @@
...
+echo ""
+
+if [ ! -f "$ORIGFILE" ] ; then
+  echo "Backing up your previous hosts file.." 
+  cp -v /etc/hosts $ORIGFILE # I like verbose file operations.  Can be less verbose if necessary.
+fi
+
+echo "Retrieving $HFILE from $HFSERVER"
+echo ""
+wget -O /tmp/$HFILE $HFSERVER/$HFILE
```
funilrys: I don't understand your `$HFSERVER/$HFILE` which means that we do
```
wget  -O /tmp/hosts.txt https://raw.githubusercontent.com/StevenBlack/hosts/master/hosts/hosts.txt
```

HostsFileUpdaterScript
```diff
@@ -0,0 +1,47 @@
...
+ORIGFILE="/etc/hosts.original"
+
+clear
+echo "-------------------------------------------------------------"
+echo "This script will update your Hosts file to the latest version of StevenBlack's Unified hosts = (adware + malware)"
+echo "Your original Hosts file will be renamed to $ORIGFILE"
+echo "-------------------------------------------------------------"
+echo ""
+
+if [ ! -f "$ORIGFILE" ] ; then
```
funilrys: What if the file already exists before we start?

rautamiekka:
I guess a Bash script this simple could be the ultimate fallback, but Bash is certainly bashed in favor of another shell like Dash by default, so a Bash-centered script will be as useless as a Python one without Python.

Looking at the script, it should work as is by just changing the shebang from `bash` to `sh`. Or better yet, `#!/usr/bin/env sh` so it'll pick `sh` from no matter where it was put into.

HostsFileUpdaterScript
ValkiryXDA: If you you read up HSFSERVER is where I have defined to get the new host file from, ill be editing that later to make it list based to choose from more sources. This is just my introduction to a real simple script that can be run at start up to ensure hosts update effectively without having to check. In short answer yes that's exactly it

HostsFileUpdaterScript
ValkiryXDA: If that file already exists if say it would be over written so the user woukd have to grab the original host before further updates.

HostsFileUpdaterScript
ValkiryXDA: If you can test it I didn't notice it during my tests

HostsFileUpdaterScript
funilrys: Understood so for now as it is, I don't think that it's gonna be merged. 
Can't wait to see your update :+1:

funilrys: That's not what you wrote ...

HostsFileUpdaterScript
funilrys: Okay so this pull request is incomplete then ... 
Can't wait to see the next update :wink:

ValkiryXDA:
Thanks rautamiekka for your suggestion to change shebang environment from bash to sh. I have applied the change.

HostsFileUpdaterScript
ValkiryXDA: No for this file it is as complete as i will get it. Hence the name  Simple Unix Script, Thats if this script does get accepted by @StevenBlack. If not ill just add it to my fork.

HostsFileUpdaterScript
ValkiryXDA: My bad i understand what you're meaning now. The script should skip if "$ORIGFILE" exists. So that way the hosts.original doesn't get overwritten.

HostsFileUpdaterScript
ValkiryXDA: If you would like to review please input in to your idea's. Dont review if you cant input or suggest.

gfyoung:
@ValkiryXDA : Have a look at #219.  How much of your file intersects with the work done there?

ValkiryXDA:
@gfyoung from as far as what i read thats a bash script that uses the python script to do the hosts file change. Mine is completely done in bash not linkung to python script

funilrys:
@ValkiryXDA So, you then need to implement everything that are implemented under the official .py

ValkiryXDA:
@funilrys i dont need to do that. Thats over complicating a simple script that functions as intended.

funilrys:
You're the one that said

> I think a simple bash script is a bit more viable since not everyone installs python or uses it for that matter. Thanks for your input dude

Which means that you assumes that people don't want to play with Python ... So you should at least to have the same functionality as [updateHostsFile.py](https://github.com/StevenBlack/hosts/blob/master/updateHostsFile.py)

ValkiryXDA:
@funilrys thats not what i said. Nor is that my intention for this script othereise i would have actually used the python script myself which i still havent done. Nor do i want to. This script serves the functionality that was intended without over complicating things and taking things out of context like you are now. Please stop in putting to this script unless you want to be constructive towards it. Thank you.

gfyoung:
@ValkiryXDA : For starters, unless you're an Arch user for example, Python is generally available in most distributions by default.  Thus, for this be merged, your script would need to be comparable to the Python script as @funilrys said.

Also, I'm reading your script, and it appears that all you're doing is downloading @StevenBlack hosts file.  The Python script does way more in terms of accessing, updating, downloading, and merging hosts file from a various number of sources.

If that's all that it does, I can just download the ZIP file of this repository, unzip it by hand and copy and paste the hosts file into the new location without ever having to touch a line of code.  Thus, unless you can make a good argument for why this repository needs a script for that, I would be very skeptical of this getting merged.

PR #369, opened 2017-07-28
Added "androidads23.adcolony.com" to data\StevenBlack\hosts

ScriptTiger:
Picked up "androidads23.adcolony.com" being called from Smule. Another AdColony node, "ads20.adcolony.com", is already added to one of the other lists. Apparently there are actually several of them, but I have only personally caught this one.


StevenBlack:
Thanks @ScriptTiger!

PR #366, opened 2017-07-22
Refactor out global settings usage in exclusions

gfyoung:
Title is self-explanatory.

Follow-up to #356.

PR #373, opened 2017-08-01
Add Python 3.5 testing to Travis

gfyoung:
This ensures that we use the most up-to-date API from the `mock` library for all supported versions of Python, which include Python 3.5.

xref #366

### DIFF
--- a/testUpdateHostsFile.py
+++ b/testUpdateHostsFile.py
@@ -46,6 +46,12 @@ class Base(unittest.TestCase):
     def sep(self):
         return "\\" if sys.platform == "win32" else "/"
 
+    def assert_called_once(self, mock_method):
+        if PY3 and sys.version_info < (3, 6):
+            self.assertEqual(mock_method.call_count, 1)
+        else:
+            mock_method.assert_called_once()
+
 
 class BaseStdout(Base):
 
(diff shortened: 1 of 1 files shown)


gfyoung:
@StevenBlack : No worries!  I can re-purpose to make it 3.5 compatible sans-`mock`.  Shouldn't take too long to do.

gfyoung:
@StevenBlack : How does it look now?

StevenBlack:
@gfyoung so sorry for the delay on this.

For some reason I have a persistent problem with `mock` on most of my macs, depending on which version of Python I'm running.

I'm really not digging this dependency.

gfyoung:
@StevenBlack : No worries!  There was no urgency to get this merged.

Unfortunately, I don't know what to say: `mock` is just like any other library.  It gets installed to `site-packages` just like anything else for Python 2.x.  I'm not sure why it would be causing issues on multiple Macs, unless you configured your Python environment the same way for all of them.

Thus, it seems to me that your `pip` is pointing to the wrong location, or Python for some reason is trying to import packages from a different location.  Have you tried installing other packages besides `mock` and not gotten that error?

StevenBlack:
OK.  Works on THIS Mac.  I'm happy 😄

PR #379, opened 2017-08-09
Refactor out global settings usage in update logic

gfyoung:
Title is self-explanatory.

Follow-up to #366.


StevenBlack:
Thanks as always, @gfyoung !

PR #384, opened 2017-08-13
Refactor out source data updating

gfyoung:
I was going to refactor `settings` out of `create_initial_file`, when I found that there was this entire block of logic in there updating the `sourcedata` field in `settings` that could be abstracted out into its own method, thereby partially addressing what I had set out to do.

Follow-up to #379.

### DIFF
--- a/testUpdateHostsFile.py
+++ b/testUpdateHostsFile.py
@@ -13,7 +13,7 @@ from updateHostsFile import (
     prompt_for_exclusions, prompt_for_move, prompt_for_flush_dns_cache,
     prompt_for_update, query_yes_no, recursive_glob, remove_old_hosts_file,
     supports_color, strip_rule, update_all_sources, update_readme_data,
-    write_data, write_opening_header)
+    update_sources_data, write_data, write_opening_header)
 
 import updateHostsFile
 import unittest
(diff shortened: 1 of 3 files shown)


StevenBlack:
Hi @ScriptTiger, can you review this in light of the changes I've made to remove .zip support from the repo?  Maybe close and resubmit because, well, removing .zip files and the attendant cleanup pollutes this PR somewhat  😄

gfyoung:
@StevenBlack : Yikes!  What happened here?  It looks like you did some force-pushing on the `master` branch, causing this PR to blow up like this.  I think in the future, it would preferable, if you did not do that, as that can make it harder for others to keep their forks synchronized.

gfyoung:
@StevenBlack : Luckily, my commits were not overlapping with your changes, so a hard reset of my branches followed by reapplying the changes did the trick.  Still, would have been nicer if there was no force-push to `master`... :smile:

gfyoung:
Also, Travis is telling me that you introduced a `flake8` error at some point, so I pushed a commit to fix that.

StevenBlack:
@gfyoung ya, I know, sorry for the hassle, I had to do it.

Expunging the .zip binaries from the repo reduced its size by about 85%, about 400 Mb.

gfyoung:
@StevenBlack : Were you trying to remove its history from the `git` repository as well?

StevenBlack:
@gfyoung yes.  Git doesn't diff binaries, which is why the repo grew so large, so quickly.  You can't remove a file and its past incarnations without also modifying history.  All the commits are there, as well as all the text-based files and their past versions, but the hashes are different now.

gfyoung:
@StevenBlack : Yep, that's what I suspected.  Thanks for clarifying!

PR #390, opened 2017-08-19
Included legacy support notice

ScriptTiger:
Included legacy support notice for archived hosts packages and CRLF to delegate legacy support out from Steven Black's main repository.


ScriptTiger:
Just introducing it for now. Later once the .zips are removed entirely from here, we can either remove the .zip column all together or replace those links with the legacy support ones.

readme_template.md
```diff
@@ -307,6 +307,8 @@ devices under a variety of operating systems.
 
 * [Unified Hosts AutoUpdate](https://github.com/ScriptTiger/Unified-Hosts-AutoUpdate "Unified Hosts AutoUpdate") (for Windows): The Unified Hosts AutUpdate package is purpose-built for this unified hosts project as well as in active development by community members. It's sophisticated enough to allow any novice the ability to install and uninstall the blacklist of their choosing to their local hosts file and keep it automatically up to date, while also being minimal enough to be able to be easily placed in a shared network location and deployed across an organization via group policies. And since it is in active development by community members, your bug reports, feature requests, and other feedback are most welcome.
 
+* [Hosts Packages](https://github.com/ScriptTiger/Hosts-Packages "Hosts-Packages"): Here hosts packages are maintained to provide ongoing support for legacy software requiring archived hosts files as sources or CRLF line endings. Updated regularly to stay synced with the latest hosts files from Steven Black's repository.
```
rautamiekka: There should be a space in `"Hosts-Packages"`, not a dash.

StevenBlack:
@ScriptTiger does this PR still apply?  Binary files (.zip) have been removed from the repo, and I think I'll stop there because the repo size is reasonable now.

ScriptTiger:
My Hosts-Packages repo downloads your hosts files, normalizes them to CRLF, and then archives them to .zips in the same directory structure as your repo to provide legacy support for people that bring up issues related to line endings or needing .zip packages. Getting .zips out of this repo is obviously critical, but isolating them to my repo where I can more easily manage their history without conflicting with the rest of the project seemed like it would make the transition smoother. After your latest history change, I'll have to re-push it if you agree with this reasoning. Otherwise it can be closed if it's felt as though such ongoing legacy support isn't needed.

PR #391, opened 2017-08-21
Refactor out global settings usage in file creation

gfyoung:
Refactors out the global `settings` variable from `create_initial_file`.

Follow-up to #384.

### DIFF
--- a/testUpdateHostsFile.py
+++ b/testUpdateHostsFile.py
@@ -6,14 +6,15 @@
 # Python script for testing updateHostFiles.py
 
 from updateHostsFile import (
-    Colors, PY3, colorize, display_exclusion_options, exclude_domain,
-    flush_dns_cache, gather_custom_exclusions, get_defaults, get_file_by_url,
-    is_valid_domain_format, matches_exclusions, move_hosts_file_into_place,
-    normalize_rule, path_join_robust, print_failure, print_success,
-    prompt_for_exclusions, prompt_for_move, prompt_for_flush_dns_cache,
-    prompt_for_update, query_yes_no, recursive_glob, remove_old_hosts_file,
-    supports_color, strip_rule, update_all_sources, update_readme_data,
-    update_sources_data, write_data, write_opening_header)
+    Colors, PY3, colorize, create_initial_file, display_exclusion_options,
+    exclude_domain, flush_dns_cache, gather_custom_exclusions, get_defaults,
+    get_file_by_url, is_valid_domain_format, matches_exclusions,
+    move_hosts_file_into_place, normalize_rule, path_join_robust,
+    print_failure, print_success, prompt_for_exclusions, prompt_for_move,
+    prompt_for_flush_dns_cache, prompt_for_update, query_yes_no,
+    recursive_glob, remove_old_hosts_file, supports_color, strip_rule,
+    update_all_sources, update_readme_data, update_sources_data, write_data,
+    write_opening_header)
 
 import updateHostsFile
 import unittest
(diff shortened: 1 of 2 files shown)


gfyoung:
@StevenBlack : friendly ping

StevenBlack:
Hi @gfyoung I'm not crazy about this one.  Here's why.

The semantics of configuration in this program work like this.

* The program has *defaults*
* The user provides *options*
* Mapping the *options* onto *defaults* gives us *settings*.

This is done explicitly, with purpose.

The variable `extensions_path` is nondescript.  Compare this with `settings["extensionspath"]` which tells us this value has a default that may have been overridden by the user.

I've been coding with these configuration semantics for years, and I would never go back.

Sensible?

gfyoung:
@StevenBlack : I'm not sure I follow your explanation here.  How am I overriding that configuration?  The purpose of this PR (and of the past several PR's) is to make the functions modular for ease of testing.  Let me try to see if I can extend what you just wrote:

* This program has *defaults*
* The user provides *options*
* Mapping the *options* onto *defaults* gives us *settings*.
* Dispatch *settings* to various helper functions that handle hosts file creation

That is all that I am trying to do.  As it stands you try to propagate a global `settings` variable across all of the functions while maintaining a very functional layout instead of OOP-format, which would make more sense if you are trying to encapsulate state (in this case `settings`).  Given that parameters from `settings` are not heavily shared, I prefer the functional layout.

As to your point about `extensions_path` being nondescript, I fail to see why assigning a variable to `settings["extensionspath"]` is problematic.  The following line:

~~~python
extensions_path = settings["extensionspath"]
~~~

indicates that this variable is assigned to a setting, which, in your words, has a default that may have been overriden by the user.

Does that make sense?  I'd be happy to discuss this further to see how we can not utilize a global `settings` variable with such a functional arrangement that at the same time doesn't destroy any semantics.

StevenBlack:
If changes are required to support testing, I'd still like the `settings` dict to be invoked there.

So in tests, where required, we could say 

    settings = get_defaults()
    settings.update(options)

... where options could be passed into the tests.  In the canonical case, options would not be necessary, but if someone wants non-standard options, then tests would still be valid in that case.

Currently tests use *magic values* like, for example:

    self.data_path = "data"
    self.host_filename = "hosts"
    self.blacklist_file = "blacklist"
    self.extensions_path = "extensions"

This is all well and good, but I don't see any benefit to eliminating the semantically solid `defaults` + `options` = `settings` just so we can then use magic values.

gfyoung:
@StevenBlack : Can you explain to me how I'm eliminating your semantics?  That's actually the main thing I don't understand from your previous commentary.

Actually, if you could comment on an example in the diff, that would be very helpful.

StevenBlack:
`settings["extensionspath"]` is crystal clear.

`extensions_path = settings["extensionspath"]` and thereafter using `extensions_path` is both spurious, and masks the value's origins as a setting.

If you want more flexibility for tests, that's great, make the configuration facet a first class element of testing.

gfyoung:
> extensions_path = settings["extensionspath"] and thereafter using extensions_path is both spurious, and masks the value's origins as a setting.

I think "spurious" is a little too harsh here.  I deliberately did this because I was using `settings["extensionspath"]` in multiple locations.  From a maintenance point of view, it is preferable to use a variable instead of the `settings["extensionspath"]`.

> If you want more flexibility for tests, that's great, make the configuration facet a first class element of testing.

To a degree it already is, as every test written is driven by settings that I set internally to the function.  But why do I need to the inflated `get_default_values()` settings when I can take a slimmed down version of it with the parameters that I already need?

gfyoung:
@StevenBlack : BTW, I should also mention that what you're saying now diverges a lot from your comments on several past PR's that I've submitted for similar refactorings.  I'm surprised that such an observation has been made quite late into the game, as I am *very close* to refactoring out the `global` settings variable at this point.

StevenBlack:
@gfyoung please consider dict `settings` as a primitive in this implementation.  Dict `settings` contains all the operational parameters for the current execution.  Dict `settings` is authoritative.  

The process control structure is named `settings`. 

The members of Dict `settings` should not be delegated to variables.  

That's my decision.  

If a `settings` dict is required by out-of-scope ancillary routines, then those routines should
reckon the `settings` in this standard fashion:

* `defaults` are provided by the program.
* `options` ***may*** be provided by the user or client.
* `settings` is the merging of `defaults` and `options`. 

Those ^^^ are the semantics of config I want.

If this means testing contains a mirror copy of the `defaults` used elsewhere, so be it.  I have no objections to that.

Eventually, testing needs to cover configuration too.  But that's outside the scope of this thread.

gfyoung:
@StevenBlack : If `settings` is to hold this much power over every function, then you should re-write the code to be class-based instead of functional.

However, this is a massive undertaking that is best left to you, as I would prefer not to go down the path of contributing nearly 10 PR's of my own time only to find out now that it grossly violated a design constraint on the code base.

ScriptTiger:
Maybe this project is getting extensive enough to warrant some docs? Maybe that could be the next big thing before anything else, just creating a docs directory with intentions and whatnot laid out. When misunderstandings like this occur in the future, the documentation can just be added to as part of the repo itself rather than searching through old issues/PRs. Setbacks should always make us stronger somehow.

StevenBlack:
@ScriptTiger Sure!  But not in this thread, OK?

ScriptTiger:
Certainly. The initial drafts would be more than deserving of their own PR, I should think. Not to mention they would be completely off topic to this. However, the more I think about it, @gfyoung does enjoy to flaunt his command of the English language. God forbid we draft something up that's not in keeping with the highest standards of modern English usage.

gfyoung:
@ScriptTiger: I do appreciate the humor, but I think you're pushing the English skills thing a little too far.😔

PR #395, opened 2017-08-21
Updated readme_template.md with "fakenews"

ScriptTiger:
Updated the "Extensions" section of the readme_template.md with the "fakenews" extension.


StevenBlack:
Good catch!

PR #424, opened 2017-10-25
Added "urldelivery.com" and "rocks.io"

ScriptTiger:
I ran across these on my network today and did some research. "urldelivery.com" is known for delivering viruses and "rocks.io" delivers data mining scripts. "rocks.io" may or may not fit into the new analytics section, but someone better at analyzing such scripts would be better suited to determine that than I.


StevenBlack:
Hi @ScriptTiger!

I'm curious, what reasons for including these two domains?

StevenBlack:
Sorry @ScriptTiger just seeing the comment now.  It's early, need coffee.

PR #425, opened 2017-10-25
Added JS based crypto currency mining sites

anudeepND:
Domains were borrowed form https://github.com/keraf/NoCoin


StevenBlack:
This is good!  Thank you @anudeepND!

Merging.

PR #426, opened 2017-10-25
Correct licenses for FadeMind host files

brarcher:
The FadeMind project lists a MIT license, however it did not create
several host files it provides. Instead, the copying.txt file
in the source:

http://hostsfile.org/Downloads/BadHosts.unx.zip

lists LGPL as the license. Without specifying the version, it
implies the latest. This is denoted by GPLv3+.


StevenBlack:
Thanks @brarcher!

PR #428, opened 2017-10-25
Remove license for UncheckyAds

brarcher:
The FadeMind lists the MIT license, however that project only hosts
the UncheckyAds host list and did not create it. Instead, that list
is based on content from unchecky.com.

I was unable to determine the license of that content, and am removing
the license to prevent someone from accidentally violating the
actual license by using the content.


brarcher:
My mistake on the wrong license for this host file. The FadeMind license threw me off. I've logged an issue against that project, in case they want to clarify what their MIT license actually covers.

Issue #430, opened 2017-10-31
Zarget - website visitors’ behaviour tracking


xetorixik:
Please add Zarget to the hosts list.
As described by Zarget: "Get detailed insights on website visitors’ behaviour".

Live example: www.sygic.com
Sygic track file: http://cdn.zarget.com/179341/476372.js

ScriptTiger:
Can anyone confirm further URLs to be added for this?
zarget.com
cdn.zarget.com

That example link also brings up an interesting case for Facebook events:
facebook.net
connect.facebook.net

Can anyone confirm if those URLs are only analytics and if they will conflict with normal Facebook usage? Preferably by testing both from a desktop and mobile setting, as apps make different connections.

StevenBlack:
Thanks @xetorixik and @ScriptTiger.  Closing.

PR #432, opened 2017-11-03
Added "zarget.com" and "bjorksta.men"

ScriptTiger:
Added "zarget.com" as per issue 430:
https://github.com/StevenBlack/hosts/issues/430

Added "bjorksta.men" as JS Miner
Moved "rocks.io" to JS Miner section


StevenBlack:
Thank you, @ScriptTiger!

PR #439, opened 2017-11-16
use `/usr/bin/env sudo` for cross-compatible sudo

stites:
Just a simple ternary operator since this will cover almost every case I can think of.

### DIFF
--- a/updateHostsFile.py
+++ b/updateHostsFile.py
@@ -35,7 +35,7 @@ else:  # Python 2
     raw_input = raw_input  # noqa
 
 # Syntactic sugar for "sudo" command in UNIX / Linux
-SUDO = "/usr/bin/sudo"
+SUDO = ["/usr/bin/env", "sudo"]
 
 
 # Project Settings
@@ -875,7 +875,7 @@ def move_hosts_file_into_place(final_file):
     if os.name == "posix":
         print("Moving the file requires administrative privileges. "
               "You might need to enter your password.")
-        if subprocess.call([SUDO, "cp", filename, "/etc/hosts"]):
+        if subprocess.call(SUDO + ["cp", filename, "/etc/hosts"]):
             print_failure("Moving the file failed.")
     elif os.name == "nt":
         print("Automatically moving the hosts file "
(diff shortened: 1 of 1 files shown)


rautamiekka:
May I suggest `/usr/bin/env sudo` instead as an universal solution ? It uses the environment to find `sudo`, no matter where it may be.

stites:
Much better, thanks @rautamiekka!

gfyoung:
@StevenBlack : Odd that this one was forgotten about.  In any case, I think this looks good to go.  Do you mind restarting the tests?

FadeMind:
@StevenBlack  please close this PR cause issues. New one waiting (fixed).

FadeMind:
@gfyoung See https://github.com/StevenBlack/hosts/pull/519/commits/ee411d82c9c00fffcc22a4c8a7496dd6304aaf3b

gfyoung:
Ah, gotcha.  I suppose this one has gone stale too...

stites:
Hmm... so I don't really understand what the difference between these two PRs is other than the fact that this one went stale. Looking at https://github.com/StevenBlack/hosts/pull/439/files and https://github.com/StevenBlack/hosts/pull/519/files the diffs look identical. Can someone enlighten me?

gfyoung:
@FadeMind explained it <a href="https://github.com/StevenBlack/hosts/pull/439#issuecomment-370165333">above</a>.  You had some linter errors that needed to be patched.  Since we couldn't fix your branch directly, @FadeMind just submitted another PR that encompassed yours with the fixes.

gfyoung:
Huh, that's odd...you're right.  I guess we didn't need to create another PR for this 😄 

cc @StevenBlack

stites:
Okay, so yeah. I'm not crazy. I'm seeing this: https://travis-ci.org/StevenBlack/hosts/jobs/303162997#L86 It would be insane to think that Travis checks intermediate commits. It also looks like @FadeMind introduced those linter errors in a previous commit on their own branch. This PR only contains one commit.

PR #441, opened 2017-11-19
New ads and tracking domains

anudeepND:


ScriptTiger:
@StevenBlack,I know I myself have added to the local data/StevenBlack/hosts in the past just due to the fact I love this project and don't like straying too far. However, do we have a local curation plan in place to deal with these? I know we have turned people away in the past that submit issues with long lists. I can personally assure you I only submit domains I personally encounter on my own network, but over time this, too, will grow to a long list and I just want to make sure we are also holding ourselves accountable.

StevenBlack:
Thanks @anudeepND!

@ScriptTiger these all look good to me.  Good point though.

anudeepND:
@StevenBlack and @ScriptTiger I test each domains before creating a PR, for example; I own a Xiaomi device so if I blacklist any legit domains i get to know if I break something. These domains were in my blacklist for a while and I haven't encountered any problems so far :)

Issue #411, opened 2017-10-04
Delay in loading web pages


user789465:
Hi, so I tried it for 3 months, with and without, and every time I have put the 39K+ lines in my host file, I can't navigate normally, it's like every 5min I have this "connection freeze" where I have to wait for 30 to 60sec before the pages load. 

I tried with lots of tab (many different web site) and it's the same, they all stop working and they are all loading the page, then all of at the same time (30 to 60sec later) they all load quickly.

I have a good computer (i7, 32GBram, SSD), I have a optical fiber internet;
I have tried this on a windows 7 vanilla.
even on VM it does the same.

without the big list in my host file, everything goes full speed, never lag, never freeze.
I don't know what is the issue

ScriptTiger:
What browser are you using?

user789465:
like I said, I tried with a non-VM Vanilla Windows 7 Updated Original.
my browser : Firefox. (the same thing happened with Chrome)
I don't use proxy. No effect on Online gaming. 
cmd.exe ping : freezes.
with all the test I did, doing a constant ping seems to give me less Freeze.
(cmd : ping -t 8.8.8.8) for example.

as for "black holes" I have no idea what this is. my computer use a normal router, 192.168.0.1 address, no firewall inside.

I'll do some more test, but I was hoping someone else had the same issue lol.
thx for your comment.

ScriptTiger:
You are just using a standard hosts file you downloaded here without any edits? Just downloaded and copy+paste? Did you copy and paste manually or use a utility?

Here is an example line from the hosts file:
0.0.0.0 lb.usemaxserver.de

"0.0.0.0" is the IP address that the domain name "lb.usemaxserver.de" resolves to. In the case of blacklisting/blocking domain names, the real IP address is replaced by a "black hole" address that does not take you to the real domain name. It is commonly referred to as a black hole because once a course is charted for a blacklisted website, it's traffic cannot escape an inevitable void of nothing.

kronflux:
I will note that when I first started using hosts block files such as MVPS years ago on Windows 7, I had an issue where when I first booted up Windows, I had to wait a good 5-10 minutes before my network connection would actually resolve web pages.
Eventually I updated my network driver, and that solved the problem for me.
Never experienced the issue again, even after clean installs of Windows 8 and Windows 10.

So you might want to check for a driver update for your network adapter.

user789465:
[...]

Issue #49, opened 2015-12-09
Use nine hostnames per line instead of one


lewisje:
If there are multiple hostnames on a line, the names after the first are treated as aliases for the first, which means that it takes less time to load in the file; also this trims file size by minimizing the number of occurrences of the redirect IP address in the file.

Although even 24 hostnames per line works in Unix-like systems (although too many names per line itself has its problems), Windows ignores any hostnames on a line after the first nine, so nine per line is ideal: http://forum.hosts-file.net/viewtopic.php?p=16438&sid=3e0ec8605c66da5a6a4bdd1bb49b5fbb#p16438

StevenBlack:
Hi @lewisje 

I've thought about this.

I often find myself eyeballing various regions of the hosts file for various reasons.

It seems much easier to scan a single column.

If we go to multiple hosts per line, I think I would keep it to 80-100-columns wide, or thereabouts, which would impose a constraint fewer than nine certainly.

Know what interests me greatly?  Metrics for the performance of host files as a function of orthogonal factors such as 

* `0.0.0.0` vs `127.0.0.1`
* How file length (number of lines) affects load and parse performance.
* The degree that multi-hosts per line helps, as seems reasonable to presume.

So far I've anecdotally seen few benefits, one way or another.  The hosts file lookup appears to be [sufficiently high in the latency stack](https://gist.github.com/jboner/2841832) to maybe not fret about?

Either way, I'm curious to know.

lewisje:
I think I should figure out how to precisely measure this, but I know that when I run `ipconfig /displaydns` on my Windows machine (to force-map the hostnames in the local DNS cache), it takes less time with multiple hostnames per line than with one, even if I suppress output (just printing the output often takes lots of time with long-running commands).

I'm thinking this suggestion is more akin to delivering a minified JS file for wide-scale Web deployment while retaining a properly spaced-out JS file for development.

Gitoffthelawn:
@StevenBlack wrote:

> Know what interests me greatly? Metrics for the performance of host files as a function of orthogonal factors such as...

>    - 0.0.0.0 vs 127.0.0.1
>    - How file length (number of lines) affects load and parse performance.
>    - The degree that multi-hosts per line helps, as seems reasonable to presume.

That will be extremely valuable information if anyone performs the testing.  I'm amazed that detailed tests have not already been publicly documented.  Cross-platform testing is essential, and will enhance the value of the data even further.

hd074:
[...]

PR #459, opened 2017-12-30
Added an option to compress the created hosts file.

stefanopini:
In particular, the compression option removes non-necessary lines (empty lines and comments) and puts multiple domains in each line.
This option should solve the issue #411 regarding the DNS client service of Windows.
A test for this option is still missing, I hope someone can work on it.

### DIFF
--- a/testUpdateHostsFile.py
+++ b/testUpdateHostsFile.py
@@ -100,6 +100,7 @@ class TestGetDefaults(Base):
                         "keepdomaincomments": False,
                         "extensionspath": "foo" + self.sep + "extensions",
                         "extensions": [],
+                        "compress": False,
                         "outputsubfolder": "",
                         "hostfilename": "hosts",
                         "targetip": "0.0.0.0",
@@ -249,7 +250,7 @@ class TestPromptForUpdate(BaseStdout, BaseMockDir):
 
     def tearDown(self):
         BaseStdout.tearDown(self)
-        BaseStdout.tearDown(self)
+        # BaseStdout.tearDown(self)
 
 
 class TestPromptForExclusions(BaseStdout):
--- a/updateHostsFile.py
+++ b/updateHostsFile.py
@@ -671,16 +671,13 @@ def compress_file(input_file, target_ip, output_file):
     lines_index = 0
     for line in input_file.readlines():
         line = line.decode("UTF-8")
-        if line.startswith(('#', '\n')):
-            continue
 
         if line.startswith(target_ip):
-            current_len = len(lines[lines_index])
-            if current_len < 128 and (current_len + len(line[7:])) < 192:
-                lines[lines_index] += line[7:-1]
+            if lines[lines_index].count(' ') < 9:
+                lines[lines_index] += ' ' + line[7:line.find('#')].strip()
             else:
                 lines[lines_index] += '\n'
-                lines.append(line[:-1])
+                lines.append(line[:line.find('#')].strip())
                 lines_index += 1
 
     for line in lines:
--- a/updateHostsFile.py
+++ b/updateHostsFile.py
@@ -648,8 +648,10 @@ def create_initial_file():
 
 def compress_file(input_file, target_ip, output_file):
     """
-    Reduce the file dimension removing non-necessary lines (empty lines and comments) and putting multiple domains in each line.
-    Reducing the number of lines of the file, the parsing under Microsoft Windows is much faster.
+    Reduce the file dimension removing non-necessary lines (empty lines and
+    comments) and putting multiple domains in each line.
+    Reducing the number of lines of the file, the parsing under Microsoft
+    Windows is much faster.
 
     Parameters
     ----------
@@ -672,8 +674,8 @@ def compress_file(input_file, target_ip, output_file):
             continue
 
         if line.startswith(target_ip):
-            l = len(lines[lines_index])
-            if l < 128 and (l + len(line[7:])) < 192:
+            current_len = len(lines[lines_index])
+            if current_len < 128 and (current_len + len(line[7:])) < 192:
                 lines[lines_index] += line[7:-1]
             else:
                 lines[lines_index] += '\n'
@@ -700,7 +702,8 @@ def remove_dups_and_excl(merge_file, exclusion_regexes, output_file=None):
     exclusion_regexes : list
         The list of regex patterns used to exclude domains.
     output_file : file
-        The file object in which the result is written. If None, the file 'settings["outputpath"]' will be created.
+        The file object in which the result is written. If None, the file
+        'settings["outputpath"]' will be created.
     """
 
     number_of_rules = settings["numberofrules"]
--- a/updateHostsFile.py
+++ b/updateHostsFile.py
@@ -62,6 +62,7 @@ def get_defaults():
         "keepdomaincomments": False,
         "extensionspath": path_join_robust(BASEDIR_PATH, "extensions"),
         "extensions": [],
+        "compress": False,
         "outputsubfolder": "",
         "hostfilename": "hosts",
         "targetip": "0.0.0.0",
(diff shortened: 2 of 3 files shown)


tbalden:
Excellent. How much smaller the file is in average?

stefanopini:
My hosts file, with all the extensions and some custom entries, passes from 1.47MB to 1.01MB, but the biggest change is in the number of lines: 7715 instead of 58143.
If it's useful, I can do some tests to see the average improvement on different scenarios.

tbalden:
30% is an excellent gain. I'm adding the hosts file into initramfs in my custom kernels, so the smaller it is, the better.
I'm curious how Linux/android would handle the parsing. I'll try it when I get some time, out of curiosity.

StevenBlack:
@stefanopini as you can see, `travis-ci` is failing... Why did you comment-out the teardown code for testing?

stefanopini:
@StevenBlack sorry, I forgot to check the code with flake8 before committing it.
There were some too-long lines and an ambiguous variable... It should be ok now.
I commented the second `tearDown` call because it was executed twice, causing an error in my Windows environment. Why was it called twice? I didn't get it.

updateHostsFile.py
```diff
@@ -630,7 +646,49 @@ def create_initial_file():
...
+    """
+
+    input_file.seek(0)  # reset file pointer
+    write_data(output_file, '\n')
+
+    lines = [target_ip]
+    lines_index = 0
+    for line in input_file.readlines():
+        line = line.decode("UTF-8")
+        if line.startswith('#') or line.startswith('\n'):
```
rautamiekka: This could be changed to
```python
if line.startswith( ('#', '\n') ):
```
for same effect with less code.

testUpdateHostsFile.py
```diff
@@ -249,7 +250,7 @@ def test_freshen_update(self, _):
 
     def tearDown(self):
         BaseStdout.tearDown(self)
-        BaseStdout.tearDown(self)
+        # BaseStdout.tearDown(self)
```
rautamiekka: Couldn't we do this like this to avoid erroring ?
```python
while 1:
    try:
        BaseStdout.tearDown(self)
    except:
        break
```
Of course, if there are any exceptions to deal with, deal with them appropriately, and don't forget to `break` the loop in any case.

ScriptTiger:
If line quantity is the issue, would "compressing" contiguous comment lines to a single line also reduce line quantity significantly enough to be acceptable without needing to remove comment lines entirely?

StevenBlack:
Hi @stefanopini thank you for this pull request.  I'll be evaluating this shortly, thank you for your patience!

updateHostsFile.py
StevenBlack: @stefanopini I think that @rautamiekka has a good point here.

StevenBlack:
@stefanopini can I ask you to please do the following before I merge?

- [ ] Update the ***Command line options*** section in the 
 in the [readme_template.md file](https://github.com/StevenBlack/hosts/blob/master/readme_template.md) which documentats command line options.
- [ ] @rautamiekka has a [good suggestion](https://github.com/StevenBlack/hosts/pull/459#pullrequestreview-86036270) for the code.

Thanks!

stefanopini:
@ScriptTiger the idea before the removal of the comments and the empty lines is that the hosts file created with this script shouldn't need to be edited manually.
In my scenario (using all the extensions and some custom rules), comment and empty lines are respectively 2271 and 2629 out of 58143 lines. Contiguous comment lines are 793 only, therefore I don't consider them remarkable. Nevertheless, since comment lines are not impacting a lot on the total number of lines, we could add an option to keep them in the future.

Thank you for suggesting to test the different like breaks, I didn't think about that. Unfortunately, the problem occurs with every line break (CR, LF, and CR+LF). Interestingly, Windows is capable to parse each of them.

I also tried the hosts files on a machine with up-to-date Windows 10 (build 1709) and the issue is still there.

tbalden:
As this is an opt-in type of feature, it's worth considering for addition even if windows will be fixed in the future. I for one would be very happy about the size gain in the initramfs scenario I'm using

updateHostsFile.py
stefanopini: Updated. Thank you

stefanopini:
@StevenBlack I updated the docs and the `startswith` code.
Do you think we're ready to merge now?

testUpdateHostsFile.py
stefanopini: Even if it should work, I think this is not necessary.
As far as I understand, the tearDown method closes the current standard output file (it was changed before calling the setUp method) and re-sets it to the system standard output.
Calling tearDown twice causes the closure of the system standard output, raising an error in the following.

updateHostsFile.py
```diff
@@ -630,7 +646,47 @@ def create_initial_file():
...
+            continue
+
+        if line.startswith(target_ip):
+            l = len(lines[lines_index])
```
StevenBlack: @stefanopini in the code below, I'm curious why you chose these particular limits for line length?  Why limit at all?

stefanopini: I was quite sure Windows had a length limit for each line of the hosts file, so I chose a value to reduce the file size without incur in errors.
But, searching the reference for that, I discovered that the limit doesn't refer to the the length of the line, but to length of the single hostname.
On the contrary, the limit that matters in our case is the number of supported hostnames for each line, fixed to 9. I found that also the issue #49 reports that.
Therefore, the generated hosts file must have at maximum 9 hostnames for each line, regardless of the line length. I've just committed an update with the correct `if` statement and little changes.
I apologize for the dumb error, I should have checked it before. Unfortunately, with my length limits, nearly every line contains less than 9 hosts, so I didn't notice the error.
I'll update the stats previously posted with the new values.

StevenBlack:
Thanks @stefanopini.  You're awesome.

user789465:
thanks for your continuing effort, if that can help people, good.

as for me, I'm really sorry, but I have no idea how to "compress" a host file.

onmyouji:
@user789465 

* install python and download the repo.
* from command prompt type "python testUpdateHostsFile.py -a -c"

tbalden:
@user789465
Please read the opening description and big things will be revealed to you!
"In particular, the compression option removes non-necessary lines (empty lines and comments) and puts multiple domains in each line."

ScriptTiger:
I was doing a bit more research into this and actually came full circle back to this repository, ranked second on Google to a hit on superuser.com, so I thought I'd share it here just in case anyone wanted to do a bit more digging into this: https://github.com/StevenBlack/hosts/issues/49. I am not actually personally affected by this, but it definitely is an interesting new feature indeed. So props to everyone that helped finally solve this and bring the solution to fruition.

user789465:
the compression seems to work a little;
is there a way to compress even more ?
(before I had 60sec delay lag, now it feels like 10 or 5sec)
the host file is about 5000 lines, maybe 100 would to the trick ?
how to make that happen ?

StevenBlack:
@user789465 Windows – always a problem it seems – limits this to 9 domains per line.

KostiantynO:
> @user789465
> 
> * install python and download the repo.
> * from command prompt type "python updateHostsFile.py -a -c"

Hello, how to compress hosts file without Python installation? I have VSCode if need.

ScriptTiger:
@KostiantynO, what's your OS?

KostiantynO:
> @KostiantynO, what's your OS?

Hello, man! Thanks for feedback!👍
  Win10 x64, I already figured out it by myself💪.
  I followed your link and downloaded Compressed.cmd file📄.

  Then just took hosts file with mouse🐁 and drag-and-dropped🖱️ it directly into Compressed.cmd file📄!

  It was great!!🎉 
  I didn't know that I can launch the bash script by dropping file onto *.cmd files!! 
  So it did comperresed my hosts from 3MB (115K lines) into 2MB (12K lines) by 9 domains per line.
  And DNS client with enabled DNS caching is now working fine, I hope 🤞🙏.
At least DNS cache does not eat 100% of CPU now.💻✨ 
That is some progress!🏆

ScriptTiger:
@KostiantynO, I'm glad that could help you. I also have another project that can do compression and update your hosts file automatically, if you're interested in that:

https://github.com/ScriptTiger/Unified-Hosts-AutoUpdate

PR #471, opened 2018-01-15
Added PowerShell support

kamilkosek:
Added a PowerShell script with some basic functions so you do not need to isntall Python.

### DIFF
--- a/Update-HostsFile.ps1
+++ b/Update-HostsFile.ps1
@@ -1,0 +1,170 @@
+﻿function Update-HostsFile
+{
+    <#
+    .SYNOPSIS
+    This CmdLet downloads the unified hosts file with optional extensions from 
+    https://github.com/StevenBlack/hosts and copies it to the correct folder on Windows machines. 
+
+    .PARAMETER UnifiedOnly
+    Downloads only the base hosts file without extensions
+
+    .PARAMETER Fakenews
+    Downloads the base hosts file with the Fakenews extension
+
+    .PARAMETER Gambling
+    Downloads the base hosts file with the Gambling extension
+
+    .PARAMETER Social
+    Downloads the base hosts file with the Social extension
+
+    .PARAMETER Porn
+    Downloads the base hosts file with the Porn extension
+
+    .PARAMETER All
+    Downloads the base hosts file with all extensions
+
+    .PARAMETER WhitelistFilePath
+    Path to a file which contains a list with domains which need to be whitelisted. The domains will
+    be cut out of the hosts file. 
+    The file must only contain the domains seperated be new line.
+
+    .PARAMETER BlackListFilePath
+    Path to a file which contains a list with domains which need to be blacklisted. The domains will
+    be appended to the end of the hosts file. 
+    The file must only contain the domains seperated be new line.
+
+    .PARAMETER BackupOldHostsFile
+    Tells the function to backup the original hosts file.
+
+    .EXAMPLE
+    C:\PS> Update-HostsFile
+    Downloads the file without extensions and writes the content into 
+    %SYSTEMROOT%\System32\etc\drivers\hosts
+
+    .EXAMPLE
+    C:\PS> Update-HostsFile -FakeNews -Social
+    Downloads the file with the social and Fakenews extensions and writes the content into 
+    %SYSTEMROOT%\System32\etc\drivers\hosts. You can combine any of the extensions.
+
+    .EXAMPLE
+    C:\PS> Update-HostsFile -Gambling -Social -WhiteListFilePath \\share\whitelist.txt 
+    Downloads the file with the social and Gambling extensions, cuts out the whitelisted domains
+    and writes the content into %SYSTEMROOT%\System32\etc\drivers\hosts. 
+
+    .EXAMPLE
+    C:\PS> Update-HostsFile -Gambling -Social -WhiteListFilePath \\share\whitelist.txt -BlackListFilePath \\share\blacklist.txt
+    Downloads the file with the social and Gambling extensions, cuts out the whitelisted domains,
+    appends the blacklisted domains and writes the content into %SYSTEMROOT%\System32\etc\drivers\hosts. 
+
+    .LINK
+    https://github.com/StevenBlack/hosts
+    #>
+    [cmdletBinding(DefaultParameterSetName='UnifiedOnly')]
+    param
+    (
+        [Switch]
+        [Parameter(ParameterSetName='UnifiedOnly',Mandatory=$false)]
+        $UnifiedOnly,
+        [Switch]
+        [Parameter(ParameterSetName='Extension',Mandatory=$false)]
+        $Fakenews,
+        [Switch]
+        [Parameter(ParameterSetName='Extension',Mandatory=$false)]
+        $Gambling,
+        [Switch]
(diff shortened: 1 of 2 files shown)


readme.md
```diff
@@ -255,6 +255,24 @@ To open a command prompt as administrator in the repository's directory, do the
...
+Set-ExecutionPolicy Bypass -Scope Process -Force; iex (iwr 'https://github.com/kamilkosek/hosts/raw/master/Update-HostsFile.ps1').Content
+```
+This command will load the script in memory and makes the CmdLet available in this session. To get an overview of available
+parameters just run 
+```
+Get-Help Update-HostsFile -Full
+```
+To update just the hosts file without extensions just run:
+```
+Update-HostsFile -BackupOldHostsFile
```
grantcarthew: This should be a separate cmdlet. Verb-Noun.

StevenBlack:
Hey @kamilkosek thank you very much for this contribution!

I *used* to be a Windows user; I ain't *ever* going back . 😈 

Can I ask the Windows users watching this, like @grantcarthew here (thanks Grant!)  to chip-in, evaluate, and make recommendations?  Thanks everyone!

Update-HostsFile.ps1
```diff
@@ -0,0 +1,170 @@
...
+    try
+    {
+        $stream = [System.IO.StreamWriter]::new( (Join-Path $env:windir 'System32\drivers\etc\hosts') )
+        $hostsArray | ForEach-Object{ $stream.WriteLine( $_ ) }
+    }
+    finally
+    {
+        $stream.close()
+    }
+    $null = ipconfig.exe /flushdns
```
gfyoung: @StevenBlack : Is this way of flushing the DNS cache still a sticky area for existing users of your hosts file?  Otherwise, this might be useful to add on the Python side too.

readme.md
```diff
@@ -255,6 +255,24 @@ To open a command prompt as administrator in the repository's directory, do the
 
 You can also refer to the "Third-Party Hosts Managers" section for further recommended solutions from third parties.
 
+## Updating hosts file on Windows using PowerShell (minimum version 3)((doesn´t require Python))
```
gfyoung: Too many parentheses on the "doesn't require Python" part.

readme.md
```diff
@@ -255,6 +255,24 @@ To open a command prompt as administrator in the repository's directory, do the
...
+
+**Windows 10**: Right click Start Button -> click on 'Windows PowerShell (Administrator)'
```
gfyoung: I would also suggest adding instructions for Windows 7 if possible (I doubt they should be super different though).  It maybe an oldie, but it's still a goodie. 😄

Update-HostsFile.ps1
```diff
@@ -0,0 +1,170 @@
...
+    %SYSTEMROOT%\System32\etc\drivers\hosts
+
+    .EXAMPLE
+    C:\PS> Update-HostsFile -FakeNews -Social
+    Downloads the file with the social and Fakenews extensions and writes the content into 
+    %SYSTEMROOT%\System32\etc\drivers\hosts. You can combine any of the extensions.
+
+    .EXAMPLE
+    C:\PS> Update-HostsFile -Gambling -Social -WhiteListFilePath \\share\whitelist.txt 
+    Downloads the file with the social and Gambling extensions, cuts out the whitelisted domains
```
gfyoung: "social" lowercase but "gambling" uppercase? (Multiple instances where you do this).

Update-HostsFile.ps1
```diff
@@ -0,0 +1,170 @@
...
+    $hostsArray = New-Object System.Collections.ArrayList  
+    $null = $hostsArray.AddRange(($hostsContent -split '\n'))
+    if(-not [string]::IsNullOrEmpty($WhitelistFilePath) )
+    {
+        
+        $whiteListDomains = Get-Content -Path $WhitelistFilePath
+        $hostsToWhiteList = New-Object System.Collections.ArrayList
+        foreach($hostsEntry in $hostsArray)
+        {
+            if($whiteListDomains.Contains($hostsEntry.Split('\s')[1]))
```
gfyoung: IINM, `.Contains` doesn't do regex matching?  I ask because exclusions on the Python-side can be regex.

StevenBlack: Thanks @gfyoung.

gfyoung:
This strikes me as something similar to what @ScriptTiger did with his own repository:

https://github.com/ScriptTiger/Unified-Hosts-AutoUpdate

My concern is that this script's functionality doesn't fully match what the Python side can do (e.g. the regex part).  Thus, despite being a Windows person, I'm not sure I'm fully on board with incorporating this into the main repository, similar to what happened with #365 on the Linux / UNIX end.

I might recommend taking this script and making it a separate repository that people can use similar to what @ScriptTiger did.

ScriptTiger:
It's completely possible to have Windows scripts do everything the Python script can, but it would require much more iterating loops to make up for functions not native in Windows. Unless further binary packages are included with it to cover the functionality gaps, performance would become the biggest issue for the longer running iterations. Which is why my script takes every possible shortcut and includes emergency cutoff functionality in the event any of these shortcuts cease working to give me time to re-script and prevent possible critical data loss.

In the end the current Python solution reigns supreme in terms of performance. So unless a native Windows solution pops up that can turn the tide there, there hasn't been any need for me to script in the gaps as long as I can take shortcuts. Also, my script is meant to run on the bare minimum Windows system with the ability to be deployed across networks, and in my experience I have run into a lot of systems that have had PowerShell locked down and disabled for security concerns, but that's just the systems I have worked with and could very likely be an unnecessary precaution in your environment. PowerShell certainly has its advantages over the powerless shell any day, but it just depends on what you're going for and will ultimately be for a niche use case like my script, which is why I maintain my own repo.

I am not really debating one way or the other here, it's up to @StevenBlack. But since @gfyoung has already been strafing spell checks and plugged me twice, I thought I'd just give my rant. @StevenBlack can blast us now, accept or reject, and close and lock the PR :P

grantcarthew:
I agree with @gfyoung, I don't think this should be part of the `hosts` repository. Anyone can make a PowerShell script.

I suggest publishing it into the PowerShell Gallery and allow people to install it from there. I published [this repo](https://github.com/grantcarthew/GCPowerShell) into the [PowerShell Gallery](https://www.powershellgallery.com/packages/GCHostsFile/0.3) which includes a cmdlet to install the block list.

Add this script into the gallery as another option for anyone interested. Just put references to the modules in the third party list.

gfyoung:
@kamilkosek : Do you have any objections to what @grantcarthew has suggested?  Otherwise, @StevenBlack I think the consensus is to close this.

Update-HostsFile.ps1
gfyoung: @grantcarthew : Any input on this BTW?  Is this command (somewhat) backwards compatible for older Windows versions?

kamilkosek: Hi , yeah it should compatible with Win 7+ , the important point is the installed PowerShell version, it must be at least PowerShell v3

StevenBlack:
Thank you for this PR, @kamilkosek.  And thank you to all who have commented here.  I've decided to close this pull request.

PR #473, opened 2018-01-18
more snooper bs

paxperscientiam:
OK, `www.benelli.angiemktg.com` leads to the same mystersnooper site I PR'd earlier. That being the case, I see no reason not to block the domain `angiemktg.com`.

### DIFF
--- a/data/StevenBlack/hosts
+++ b/data/StevenBlack/hosts
@@ -6,6 +6,7 @@
 127.0.0.1 2468.go2cloud.org
 127.0.0.1 adsmws.cloudapp.net
 127.0.0.1 androidads23.adcolony.com
+127.0.0.1 angiemktg.com
 127.0.0.1 annualconsumersurvey.com
 127.0.0.1 apps.id.net
 127.0.0.1 breaksurvey.com


ScriptTiger:
We're going to run into a problem with this one. "angiemktg.com" (192.241.183.165) is some kind of Craigslist spammer hosted on a DigitalOcean shared host (msvm.adns1.net/192.241.183.165). They are spamming endless subdomains (www.fillthevan.angiemktg.com, Flame.angiemktg.com, bukashi.angiemktg.com, loaficcialgears.angiemktg.com, and the list goes on and on and on), which can only be blocked through the use of wild cards or inheriting a ban from a parent domain, neither of which are supported in basic hosts files. If you block "angiemktg.com", it still must be a separate host entry from "www.benelli.angiemktg.com" and all of it's countless other domains. However, this is definitely a good entry to have commented somehow though for those people converting this list to other formats that do support more dynamic blocking functionality. I would suggest moving your entry to a new "Subdomain Spam" or "Watch List" section or something to that effect and commenting it like so:

    # angiemktg.com # Subdomain Spammer

@StevenBlack will ultimately have to make a suggestion on this one though. I think if we have a section like this well documented, it's helpful for people using the data in other formats so they know to check such a section to see if they can make any additional dynamic changes on whatever platform they're working with. If we go that route, we should also add a note to the README to make such people aware.

@Gitoffthelawn, maybe your list can find some utility here if you are still curating one that is difficult to pin to a category.

StevenBlack:
Thanks @paxperscientiam I'll merge this now because I think this domain should be blocked.

PR #488, opened 2018-02-03
Added and loaded a new spam domain section

ScriptTiger:
Added and loaded a new spam domain section. These domains have have been found to be the root domains of spam either in forums, e-mail, or other online communications. The ones loaded so far contain social engineering attempts, but the intent of the section may vary over time.


StevenBlack:

PR #490, opened 2018-02-07
Added commented wildcards for spam domains to support third-party, downstream uses

ScriptTiger:
Added commented wildcards for spam domains so that they may be more easily scripted for automated integration into other platforms by finding and replacing lines that start with "#.*", while also making it easy to simply uncomment them if wildcards are accepted by their software.


StevenBlack:
Revisiting this....  no harm in this, I suppose.  And it enables downstream functionality.

PR #491, opened 2018-02-09
updateHostsFile.py: add ipv6 additionals

FadeMind:
@StevenBlack this is default file in Manjaro Linux. This pull adding extra ipv6 fields data.
screenshot_20180209_192039


StevenBlack:
Thanks Tomasz @FadeMind.  Merging.

PR #492, opened 2018-02-10
Spotify-Ad-free init

FadeMind:
Implemented https://github.com/StevenBlack/hosts/issues/483

- removed `SpotifyAds` sources
- added `data/Spotify-Ad-free/update.json` init file

See: https://github.com/FadeMind/hosts.extras/commit/e5702b4e40edd63f4c5a53322e6cd3fe65862230

@StevenBlack

PR #497, opened 2018-02-12
hosts url update

Laicure:


Laicure:
oh sorry, dup with #492

PR #503, opened 2018-02-16
Add Option to Minimise hosts File

ankitpati:
This option (`--minimise`, `-m`) differs from compress (`--compress`, `-c`) in that it keeps each host on a different line, while still removing unnecessary newlines and comments.

This is necessary because many implementations of URL blockers that rely on `hosts` files do not conform to the standard which allows multiple hosts on a single line.

### DIFF
--- a/readme_template.md
+++ b/readme_template.md
@@ -109,6 +109,11 @@ ignoring non-necessary lines (empty lines and comments) and putting multiple
 domains in each line. Reducing the number of lines of the hosts file improves 
 the performances under Windows (with DNS Client service enabled).
 
+`--minimise`, or `-m`: `false` (default) or `true`, like `--compress`, but puts
+each domain on a separate line. This is necessary because many implementations
+of URL blockers that rely on `hosts` files do not conform to the standard which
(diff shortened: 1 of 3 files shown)


StevenBlack:
Thanks Ankit @ankitpati – very nice.

But... can I ask you to update docs, ind the `Command Line Options` section of the [readme_template.md](https://github.com/StevenBlack/hosts/blob/master/readme_template.md#command-line-options)?

ankitpati:
You’re welcome, and thank you for reviewing the PR.

I updated `readme_template.md`, but I `grep`ped through the whole repo, and cannot find the “docs” which you mention. Essentially, I matched every mention of `compress`, in whatever I could find, with a mention of `minimise`.

I fixed the failing tests, but I feel `test_get_defaults` is overly restrictive and only serves to duplicate work whenever a default option is changed or a new one is added.

StevenBlack:
Ankit @ankitpati thanks!

`readme_template.md` is the only place where the change is required; you got it.

ankitpati:
Thank you for the merge.

Issue #444, opened 2017-11-27
Updating hosts data should not clobber existing hosts data if updating source times out or fails


StevenBlack:
Currently if a hosts update fails for any reason, we delete the old pre-existing source's hosts file.

StevenBlack:
Can't reproduce now.  Weird.

Closing.

StevenBlack:
Reopening this issue.

I was playing with AWS Cloud9, and came across this:

If `someonewhocares.org` fails, for example, I still expect to see about 43,000 hosts (our latest tally for the base hosts file).

This is a bug.

![2017-12-03_00-39-03](https://user-images.githubusercontent.com/80144/33522778-b709b60a-d7c2-11e7-809e-dc4a487840ee.jpg)

Issue #418, opened 2017-10-13
updating freezes -- should we timeout?


stolk:
When updating, it freezes for a long time because of malwaredomainlist.com

```
$ python3 updateHostsFile.py 
Do you want to update all data sources? [Y/n] 
Updating source extensions/fakenews from https://raw.githubusercontent.com/marktron/fakenews/master/fakenews
Updating source extensions/social from https://raw.githubusercontent.com/Sinfonietta/hostfiles/master/social-hosts
Updating source extensions/porn/sinfonietta from https://raw.githubusercontent.com/Sinfonietta/hostfiles/master/pornography-hosts
Updating source extensions/porn/sinfonietta-snuff from https://raw.githubusercontent.com/Sinfonietta/hostfiles/master/snuff-hosts
Updating source extensions/porn/clefspeare13 from https://raw.githubusercontent.com/Clefspeare13/pornhosts/master/0.0.0.0/hosts
Updating source extensions/gambling from https://raw.githubusercontent.com/Sinfonietta/hostfiles/master/gambling-hosts
Updating source data/yoyo.org from https://pgl.yoyo.org/adservers/serverlist.php?hostformat=hosts&mimetype=plaintext&useip=0.0.0.0
Updating source data/mvps.org from http://winhelp2002.mvps.org/hosts.txt
Updating source data/someonewhocares.org from http://someonewhocares.org/hosts/zero/hosts
Updating source data/tyzbit from https://raw.githubusercontent.com/tyzbit/hosts/master/data/tyzbit/hosts
Updating source data/StevenBlack from https://raw.githubusercontent.com/StevenBlack/hosts/master/data/StevenBlack/hosts
Updating source data/Badd-Boyz-Hosts from https://raw.githubusercontent.com/mitchellkrogza/Badd-Boyz-Hosts/master/hosts
Updating source data/add.Dead from https://raw.githubusercontent.com/FadeMind/hosts.extras/master/add.Dead/hosts
Updating source data/malwaredomainlist.com from https://www.malwaredomainlist.com/hostslist/hosts.txt
Problem getting file:  https://www.malwaredomainlist.com/hostslist/hosts.txt
Error in updating source:  https://www.malwaredomainlist.com/hostslist/hosts.txt
Updating source data/add.2o7Net from https://raw.githubusercontent.com/FadeMind/hosts.extras/master/add.2o7Net/hosts
[...]

PR #504, opened 2018-02-17
malwaredomainlist.com is available over HTTPS

ankitpati:
It is currently being fetched over [HTTP](http://www.malwaredomainlist.com/hostslist/hosts.txt "Malware Domain List over HTTP"), although it is readily available over [HTTPS](https://www.malwaredomainlist.com/hostslist/hosts.txt "Malware Domain List over HTTPS").

### DIFF
--- a/data/malwaredomainlist.com/update.json
+++ b/data/malwaredomainlist.com/update.json
@@ -1,9 +1,9 @@
 {
     "name": "Malware Domain List",
     "description": "Malware Domain List is a non-commercial community project.",
-    "homeurl": "http://www.malwaredomainlist.com/",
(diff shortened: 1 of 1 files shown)


funilrys:
@ankitpati Please take a look at #418 👍

StevenBlack:
@ankitpati available... until it isn't.  We've had issues with this provider before.

ankitpati:
@StevenBlack Would you be more willing to accept this PR if I can implement a timeout logic, as mentioned in #418?

StevenBlack:
@ankitpati if you could, that would be great!  See also #444 whereby, in some circumstances, the old pre-existing hosts file is not preserved.

PR #516, opened 2018-02-26
Removed Cryptominer Domains

ScriptTiger:
Before this PR can be approved, the following list of domains should be discussed as they were not found on @ZeroDot1's list. I've also included some comments.

    # JS based cryptocurrency mining sites
    127.0.0.1 joyreactor.cc
    127.0.0.1 coinminerz.com
    127.0.0.1 cryptonight.wasm # Probably mistaken as a URL, but this is actually the name of a WebAssembly binary package used by miners
    127.0.0.1 www.deepc.cc # deepc.cc is on CoinBlockerList
    127.0.0.1 cryptocoinjs.com
    127.0.0.1 www.cryptocoinjs.com
    127.0.0.1 www.bjorksta.men # bjorksta.men is on CoinBlockerList

### DIFF
--- a/data/StevenBlack/hosts
+++ b/data/StevenBlack/hosts
@@ -217,49 +217,6 @@
 127.0.0.1 yashin.alphonso.tv
 127.0.0.1 zico.alphonso.tv
 
-# JS based cryptocurrency mining sites
-127.0.0.1 server.jsecoin.com
-127.0.0.1 jsecoin.com
-127.0.0.1 www.jsecoin.com
-127.0.0.1 load.jsecoin.com
-127.0.0.1 listat.biz
-127.0.0.1 mataharirama.xyz
-127.0.0.1 static.reasedoper.pw
-127.0.0.1 lmodr.biz
-127.0.0.1 minecrunch.co
-127.0.0.1 minero.pw
-127.0.0.1 ppoi.org
-127.0.0.1 kissdoujin.com
-127.0.0.1 joyreactor.cc
-127.0.0.1 anime.reactor.cc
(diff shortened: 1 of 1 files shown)


paxperscientiam:
With regard to entries that found in a different source, is it worth the effort if the compile process removes the duplicates? I'm not even 100% sure duplicates are removed tbh.

StevenBlack:
@paxperscientiam duplicates are removed, yes.

ScriptTiger:
Duplicates in this case should be removed if we have a dedicated curator to handle their curation. For instance, if rocks.io drops off the grid over night, Steven isn't checking all those domains to see if they are all still active, still have the same scripts available, still under the same management, etc. And that is not a point against Steven, it's just impossible for any one person to give that much time and attention to as many domains as we list in this repo. ZeroDot1 works hard in that very focused area to provide a high quality product so we don't have to worry about it and our end result stays as efficient as possible with only active and relevant domains.

ZeroDot1:
@ScriptTiger @StevenBlack @dnmTX 
https://github.com/ZeroDot1/CoinBlockerLists/commit/a5f79c04d8d0b674cdf767f40b63776b589c0340
https://github.com/ZeroDot1/CoinBlockerLists/issues/120
With the, Last Update I added a white list to improve the use of the CoinBlockerLists, I hope it's a good thing that helps all users to use the CoinBlockerLists well.

StevenBlack:
Thanks as usual, @ScriptTiger.

dnmTX:
Oops....forgot to thank.....THANK YOU @ScriptTiger @StevenBlack @ZeroDot1

PR #517, opened 2018-02-26
phish

paxperscientiam:
Both of these ".win" sites serve up rotten phish (adobe flash).


StevenBlack:
Thanks Chris @paxperscientiam.  Merging.

PR #518, opened 2018-02-26
Added crypto mining sites

anudeepND:
https://www.bleepingcomputer.com/news/security/ad-network-uses-dga-algorithm-to-bypass-ad-blockers-and-deploy-in-browser-miners/


anudeepND:
These domains were borrowed from https://github.com/Yhonay/antipopads/blob/master/hosts
Thanks @Yhonay

dnmTX:
@anudeepND  so judging by the article you posted all the domains that are being generated are used once and then disposed,so is there a point to add that many domains if they are already been used(by my guess) and will never be used again?I just pinged some of them and got "request time out".Unless you tell me that these domains are for future use i don't really see a point of adding them.

anudeepND:
@dnmTX you're right, let @StevenBlack decide whether to close or merge this PR

anudeepND:
Closing...

PR #525, opened 2018-03-03
add OpenBSD support

MitchWeaver:
if check on platform.system(), as OpenBSD has moved to its own 'sudo' alternative: `doas`


StevenBlack:
Thanks Mitch @MitchWeaver !  Beauty.

PR #535, opened 2018-03-09
Remove Superfluous Check for Zero-Length List

ankitpati:
Checking for zero-length is not necessary here as iterating over a zero-length list is perfectly valid, and produces exactly zero iterations, matching the intended behaviour without an extra check.


StevenBlack:
@ankitpati nice!  Merging.

PR #557, opened 2018-04-03
Blocked ads on Hotstar streaming apps

anudeepND:
Google play link: [Hotstar](https://play.google.com/store/apps/details?id=in.startv.hotstar)
Website: [Hotstar](http://hotstar.com)

Hotstar is a popular streaming service with over 100 million downloads on Google play and over 300 million+ users . 
This PR will:
1. Blocks all kinds of video/banner ads and pixel tracking in both app and website.
2. Doesn't cause any slowdowns.
3. Doesn't block any legitimate service.
4. Does not cause any delay while loading the videos.

### DIFF
--- a/data/StevenBlack/hosts
+++ b/data/StevenBlack/hosts
@@ -421,6 +421,30 @@
 0.0.0.0 vidutils.taboola.com
 0.0.0.0 wf.taboola.com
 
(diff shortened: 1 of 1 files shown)


anudeepND:
@StevenBlack they also have Canadian version, `http://ca.hotstar.com` can you check whether it's ad free?

PR #563, opened 2018-04-09
Remove sentry.io from StevenBlack list

jooeycheng:
https://sentry.io is an error tracking tool, not an ad site.

I believe it was added by mistake from https://github.com/StevenBlack/hosts/pull/557

### DIFF
--- a/data/StevenBlack/hosts
+++ b/data/StevenBlack/hosts
@@ -435,7 +435,6 @@
 0.0.0.0 cdn.segment.com
 0.0.0.0 cdn.mxpnl.com
 0.0.0.0 js-agent.newrelic.com
-0.0.0.0 sentry.io
 0.0.0.0 sb.scorecardresearch.com
 0.0.0.0 in-star.videoplaza.tv
 0.0.0.0 segment.hotstar.com


anudeepND:
@StevenBlack and @jooeycheng Regardless of the function, it's a telemetry site.

StevenBlack:
@jooeycheng and @anudeepND looking at the sentry.io site, it appears to function much like crashlytics, which we also block.

I'm OK with blocking this on principle, but I may be tainted because crashlytics is so pervaisive.

Which is to say, I can be convinced either way, to the degree that sentry.io is somehow less scummy than crashlytics.  I just don't know.

anudeepND:
@StevenBlack I agree, but the basic error report contains lots of device data and other logs, so it should be blocked.

jooeycheng:
@StevenBlack @anudeepND got it, understand your point.
This affects me bcos i use sentry at work. I shall add it to my whitelist then. 👍

StevenBlack:
Thanks @jooeycheng because it's interesting either way.

It's a tough call.  In some ways, Crashlytics poisons the well.  Looking at the "sensitive data" section of docs.sentry.io, that seems to have sensible defaults, but it's configurable....

Issue #576, opened 2018-04-19
analytics.publitas.com


xetorixik:
Please add analytics.publitas.com

"This is why we’ve made it possible to track the original source of your traffic (e.g. an affiliate site or one of your email campaigns). We do this by automatically taking the URL parameters (e.g. UTM tags) from your publication’s URL and adding them to all the links in your publication."

Thanks.

PR #577, opened 2018-04-19
Analytics.publitas.com

xetorixik:
Adding analytics.publitas.com as it is used as data Analytics.
Discussion > https://github.com/StevenBlack/hosts/issues/576


StevenBlack:
Thanks @xetorixik!  Merging.

Issue #585, opened 2018-04-22
Skillshare Block


tooxo:
Dont know about the PC but on android the streaming of the skillshare app gets blocked.

StevenBlack:
Hi @tooxo do you have any deets?  At minimum we need to know, which domains the Skillshare app depends upon.

tooxo:
dont know, is there a wireshark for android?

anudeepND:
@tooxo Try whitelisting `edge.api.brightcove.com` and `players.brightcove.net`

Rikk:
Packet Capture is a simple app for capturing packets; it can decode https traffic too: https://play.google.com/store/apps/details?id=app.greyshirts.sslcapture

tobibeer:
Thought it was a flash issue and then realized my hosts file broke SkillShare. Had to whitelist **f1.media.brightcove.com**.

PR #586, opened 2018-04-22
Remove edge.api.brightcove.com and players.brightcove.net

anudeepND:
https://github.com/StevenBlack/hosts/issues/585

### DIFF
--- a/data/StevenBlack/hosts
+++ b/data/StevenBlack/hosts
@@ -80,7 +80,6 @@
 127.0.0.1 files.ctnet4.space
 127.0.0.1 apps3.cointraffic.io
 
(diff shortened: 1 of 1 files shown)


anudeepND:
@StevenBlack I don't think these domains serve ads or  does any kind of tracking, it's just a online video player solution. 
https://www.brightcove.com/en/
https://en.wikipedia.org/wiki/Brightcove

StevenBlack:
@sorry @anudeepND !  I already pushed ba4d54f1b7269d960e1d22d07a90e5d837e16df0.

Even though this PR landed first, but I didn't see it. 😦

Issue #597, opened 2018-05-01
Deprecation warning when running testUpdateHostsFile.py


onmyouji:
Hi,

When running testUpdateHostsFile.py, I got this error message

> .......C:\Program Files\Python36\lib\site-packages\bs4\builder\_lxml.py:250: DeprecationWarning: inspect.getargspec() is deprecated, use inspect.signature() or inspect.getfullargspec()
>   self.parser.feed(markup)

Does anyone else get this error message?

StevenBlack:
Hi @onmyouji yes, I get same on MacOS.

funilrys:
Will take the time this weekend (or part of the next week) to fix this along with #528 Steven @StevenBlack.

Cheers,
Nissar

funilrys:
Hello @onmyouji, 
I mentioned it into #602 but for clarification for the others who think that it is a local issue, please report to #520 description:

> Please note that I have to submit an issue to upstream but if you see the following or related when runing the tests under `Python >= 3` it's not a local issue. 
>```
>/home/travis/miniconda3/envs/hosts/lib/python3.5/site-packages/bs4/builder/_lxml.py:250: >DeprecationWarning: inspect.getargspec() is deprecated, use inspect.signature() instead
>```

For now, the maintainers of `lxml` and especially `bs4` know that the following warning exists but consider that as not a priority to fix yet. 

I personally going to contact one of the maintainers to hear more from their part as that warning is full of sense but until then, we keep the current version of the scripts unless someone has another idea or Steven @StevenBlack decide to drop BeautifulSoup.

Have a nice day/night.

Cheers,
Nissar.

onmyouji:
Thanks @funilrys

PR #602, opened 2018-05-05
Deletion of deprecation warning (Fix of #597)

funilrys:
Please note that this patch do the following:
 * Deletion of lxml from dependencies
 * Usage of `html.parser` (Python built-in parser) instead of `lxml` as the parser used by BeautifulSoup

Explanation:
  As the maintainer of the `lxml` package is aware of that warning but it still not fixed, I prefer to use `html.parser` which is the built-in Python HTML parser.
  This way, we avoid the `lxml` deprecation message mentioned by #597.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,2 @@
-lxml==4.1.1
-beautifulsoup4==4.6.0
 mock==2.0.0
+beautifulsoup4==4.6.0
--- a/requirements_python2.txt
+++ b/requirements_python2.txt
@@ -1,3 +1,2 @@
 mock==2.0.0
-lxml==4.1.1
 beautifulsoup4==4.6.0
--- a/updateHostsFile.py
+++ b/updateHostsFile.py
@@ -23,7 +23,6 @@ import tempfile
 import time
 from glob import glob
 
-import lxml  # noqa: F401
 from bs4 import BeautifulSoup
 
 # Detecting Python 3 for version-dependent implementations
@@ -1229,7 +1228,7 @@ def get_file_by_url(url):
 
     try:
         f = urlopen(url)
-        soup = BeautifulSoup(f.read(), 'lxml').get_text()
+        soup = BeautifulSoup(f.read(), 'html.parser').get_text()
         return '\n'.join(list(map(domain_to_idna, soup.split('\n'))))
     except Exception:
         print("Problem getting file: ", url)


StevenBlack:
Thanks Nissar @funilrys!

Using this branch, the size of the default hosts file decreases from 54,398 to 33,117 hosts.

Is this what you see as well?

What's up with that?

funilrys:
Side note: I also tried the `html5lib` parser and it ends up with the same issue.

StevenBlack:
Why don't we ask @mitchellkrogza to tone-down the volume on his file header?  

Or we could just use Python's native parser.  This isn't rocket science, after all...

funilrys:
@StevenBlack He's away from keyboard until Thursday normally but we will talk about that internally :+1: 

This patch was using the native Python's parser :wink: :joy_cat: 

Actually I'm more surprised that someonewhocares was generated correctly desperate the fact that some non-HTML tag are present ...

That means that that issue is because we are do not have an opening HTML tag ... Not because `</_` is not a valid HTML tag as I mentioned previously... :thinking:

funilrys:
Also Steven @StevenBlack, sorry for coming back to this but, if we do change the header at BaddBoyzz, if tomorrow another list have the same issue we can not oblige the maintainer of that list to change their list in order to apply our desired format :thinking: 

So basically using the Python native parser will give us much more headache that this warning showing up at the beginning of each test (for now) ... 

What's your opinion on this?

Have a nice day/night.

StevenBlack:
I haven't looked at this closely yet, but it seems straightforward to me.  Ignore all lines that commence with a comment character.  End of story.  What we're doing now is flawed because, can you think of a more straightforward and easy parsing rule?

funilrys:
To understand, we can't - unless we create our own system - because in order to auto-detect the encoding of the file we are downloading in order to avoid issues we use https://github.com/funilrys/hosts/blob/3bd78f523d95861e43d8508f0ee322f15d0c6034/updateHostsFile.py#L1231 Which means, we parse the whole downloaded file into BeautifulSoup which returns us the text in UTF-8.

funilrys:
Actually, there is a solution but it's really bad ...

My idea is to escape all commented line to get the original line.

But will search a more deep solution ...

So explained with code :

```python
from html import escape


to_test = {
    "<Hello world>": "&lt;Hello world&gt;",
    "Steven & Nissar": "Steven &amp; Nissar",
    "StevenBlack/hosts": "StevenBlack/hosts",
    "</Hello world>": "&lt;/Hello world&gt;",
}

for test, expected in to_test.items():
    actual = escape(test)
    assert expected == actual, "%s != %s" % (test, actual)

    print("To replace: %s" % test)
    print("Expected: %s" % expected)
    print("Actual: %s\n" % actual)
```

Gives us 
```
Line: <Hello world>
Expected: &lt;Hello world&gt;
Actual: &lt;Hello world&gt;

Line: Steven & Nissar
Expected: Steven &amp; Nissar
Actual: Steven &amp; Nissar

Line: StevenBlack/hosts
Expected: StevenBlack/hosts
Actual: StevenBlack/hosts

Line: </Hello world>
Expected: &lt;/Hello world&gt;
Actual: &lt;/Hello world&gt;
```

PR #610, opened 2018-05-12
Adservice.google.nl

xetorixik:
Several country based ad sites are already blocked like adservice.google.com, adservice.google.com.au.
Found  a new one: adservice.google.nl

Please add.
Thanks.


StevenBlack:
Thanks @xetorixik!  Merging.

PR #623, opened 2018-05-19
Rollbar Error Tracking and Crash Reporting

ScriptTiger:
Hey, I'd like a second opinion on this before it gets approved. I caught it being called from a Chromecast and it appears to be an error tracking and crash reporting service. Obviously it includes analytics of some kind in order to do the debugging, but I am not sure if others will think this counts as analytics or not. If anyone else has a Chromecast and can test blocking this, as well, and let me know the results, that would also be most appreciative just to make sure it's not a necessary service. It is a telemetry service and I know we're in agreement about the Windows telemetry services, so I am assuming this would count.

### DIFF
--- a/data/StevenBlack/hosts
+++ b/data/StevenBlack/hosts
@@ -620,4 +620,10 @@
 0.0.0.0 token.rubiconproject.com
 0.0.0.0 fc.yahoo.com
 0.0.0.0 pr-bh.ybp.yahoo.com
+
+# Error tracking and crash reporting
+rollbar.com
+www.rollbar.com
+api.rollbar.com
+
 #=====================================


StevenBlack:
LGTM!  Merging.

dnmTX:
Steve @StevenBlack  this one doesn't look like got merged,all 3 domains are missing from latest Release 1.4.15.

ScriptTiger:
@dnmTX, I think @StevenBlack may have run the script on his local repo before he merged the changes. It looks good on GitHub the merge was successful but the resultant hosts files from the script just don't include the updates because the script was run first. They should show up on the next run though, so it's no biggie.

ScriptTiger:
OMG, @dnmTX, you are absolutely right, there is a massive problem and I am an idiot, apologies. I will make a new PR. I have been switching between domain list formats for my other projects and completely screwed that up.

PR #627, opened 2018-05-23
Doc / Check for Centos restart

### LINKED ISSUE #622: Directions for RHEL/CentOS
According to [this page](https://support.rackspace.com/how-to/centos-hostname-change/) the way to re-read the hosts file on the above OSes is

```
$ sudo /etc/init.d/network restart
```

Can we get this added to the readme?

gfyoung:
Closes #622.

### DIFF
--- a/updateHostsFile.py
+++ b/updateHostsFile.py
@@ -1063,6 +1063,15 @@ def flush_dns_cache():
                 else:
                     print_success(nscd_msg.format(result="succeeded"))
 
+        centos_file = "/etc/init.d/network"
+        centos_msg = "Flushing the DNS cache by restarting network {result}"
+
+        if os.path.isfile(centos_file):
+            if subprocess.call(SUDO + [centos_file, "restart"]):
+                print_failure(centos_msg.format(result="failed"))
+            else:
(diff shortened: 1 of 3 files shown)


StevenBlack:
Thanks @gfyoung !

Merging.

updateHostsFile.py
```diff
@@ -1063,6 +1063,15 @@ def flush_dns_cache():
                 else:
                     print_success(nscd_msg.format(result="succeeded"))
 
+        centos_file = "/etc/init.d/network"
```
funilrys: May I ask if we are sure that this file is only present into CentOS ?

gfyoung: I'm not aware of any other distros at the moment.

funilrys: `Redhat`, `Fedora`, `Suse` and `OpenSuse Linux` ?

 I'm pretty sure there was (and maybe still is) a `/etc/init.d/network` into those distros.

gfyoung: Ah, that's right.  In the docs, it's referred to as `RHEL / Centos` (though obviously such a variable name isn't possible in Python).

PR #628, opened 2018-05-23
Corrected format of Rollbar domains

ScriptTiger:
The black hole address was absent, causing the script to not read them.


StevenBlack:
I missed  detail as well 😄 
ing.
Merg

PR #632, opened 2018-05-24
Android devices ship with pre-installed malware

anudeepND:
A detailed report can be found [here](https://blog.avast.com/android-devices-ship-with-pre-installed-malware)


anupritaisno1:
Thanks for the request but it seems users should actually be careful about what they are buying

Also these weren't officially certified by Google. Google certified devices usually have this https://source.android.com/security/verifiedboot/verified-boot and that would want the user if this was happening

StevenBlack:
Thank you @anudeepND.

Merging.

Issue #493, opened 2018-02-11
Chinese Information Gathering


ScriptTiger:
Does anyone know any good sources to get all the Chinese information gathering domains? I have been doing a bit of research on some IoT devices and apparently any Chinese software that slips in talks back to Baidu, Tencent, etc. Effected devices include TVs, tablets, and anything else they can get into your home apparently, so you might want to think twice before thinking you got a deal on some cheap Chinese stuff online. I have also been doing some testing with some rather popular Android apps that are loaded with such Chinese software, playing around with an old junker phone I can afford to have the IMEI of stolen or hardware reset a few times. For anyone that's not familiar, if you are turned off by things like Google Analytics...well...you might just be shocked by some of the things the Chinese are doing in their apps to say the least.

Atavic:
Maybe is it useful?
EasyList Lite China: https://github.com/cjx82630/cjxlist/

ScriptTiger:
>! Title: CJX's Annoyance List
! Supplement for "EasyList China+EasyList" & "EasyPrivacy"
! Removed Annoyances, Self-promotion & Privacy Protection on Chinese Websites
! Last modified: 2018/02/12 14:21 +0800
! Expires: 4 days (update frequency)
! License: http://creativecommons.org/licenses/by/3.0/
! Email: cjxlist@gmail.com
! Homepage: http://abpchina.org/forum/forum.php?mod=viewthread&tid=29667

It looks like they have some good data, but the formatting is a bit off for this repo and they have a lot of Chinese characters which might throw off the script like Cyrillic has in the past. I'll submit an issue with them and see if they would be willing to curate their data in hosts format, as well. Steven will have to do his review for it to join this repo, but I think it would at least benefit their follows in any case.

Atavic:
I stumbled upon some IP lists in the past, maybe related to shadowsocks, XX-net or GoAgent...

Now I found that many use a [gwflist2pac](https://github.com/search?utf8=%E2%9C%93&q=gfwlist2pac) that translates a base64 list into an usable PAC File, e.g. [this outdated output](https://github.com/KyonLi/ss-pac/blob/master/BlackList.txt).

ScriptTiger:
Those blacklists (as referenced here: https://github.com/gfwlist/gfwlist) are lists of domains that are blacklisted by the Chinese Government. Doing a quick search, you'll notice every domain of Facebook is on it, but no mention of Tencent or QQ or Baidu.com.

[...]

PR #648, opened 2018-06-07
UC apps telemetry 

anudeepND:
Tracking domains which are used in UC apps like UC browser, UC news etc.  #493


StevenBlack:
Thank you, @anudeepND !

Merging.

Issue #634, opened 2018-05-27
[redirect/tracking/ads] new domain(s)


dnmTX:
**click.linksynergy.com**  redirect and most likely tracking as well  so it belongs here

**ogp.me** (Open Graph Protocol) Used to deliver popup ads,wider testing is needed on this one,might break some Facebook functions,already blocked it on my end will update if i notice something weird.
**Update**: Yahoo is using it for ads.....tbc

**ero-advertising.com** domain that delivers ads(porn sites mostly)
_____________________________________________________________________________________________________________
Steven @StevenBlack  let's keep the issue open for now so i can keep adding if any,if there is nothing to add i'll  close it myself.Thank you.

dnmTX:
Steven @StevenBlack is it possible(on your end) when all the hosts lists are arranged into one a line to be inserted at the beginning and the end of each individual lists.For example:
# KADhosts Start
domain.com
domain.com
domain.com
# KADhosts End
It would really help me remove things that i don't need(as we discussed yesterday with you),i managed to do it now but if you change the arrangement it will be a mess on my end

P.S. If it's too much trouble maybe just something to indicate the end will be sufficient also.
P.S.P.S. You didn't answer my question @ZeroDot1's repo,so i'm posting it here 😜

StevenBlack:
@dnmTX sure, we can do that.

Is `# {source name} Start` and `# {source name} End` sufficient for your needs?

Asking because, for parsing purposes, I know that Is `# start {source name}` and `# end {source name}` seems easier.  

Also, I suggest normalizing the case of the `{source name}` to lowercase.

(I wish you had opened a new issue, rather than appending to a non-related issue.  I'll rename this and delete the first comment).

dnmTX:
Steven @StevenBlack  many many thanks(i mean it).Lowercase...uppercase....doesn't really matter,whichever is easiest for you,either way i have to parse the whole thing in order "sed" to recognize it.**THANK YOU**

I realize that lately me opening a "issue" is me being picky,so i just trying not to and looking for a way to fix things on my end(which is also educational).Sorry about that and if you'd like i'll open a issue for this one.

StevenBlack:
Turns out, I can't delete the top level comment to clean up this mess so, yes please, create a new issue.

Closing.

PR #651, opened 2018-06-09
Introduction of `Start xx` and `End xxx`

### LINKED ISSUE #644: [change request] hosts lists arrangement
Steven @StevenBlack is it possible(on your end) when all the hosts lists are arranged into one a line to be inserted at the beginning and the end of each individual lists.For example:
**# KADhosts Start**
domain.com
domain.com
domain.com
**# KADhosts End**
It would really help me remove things that i don't need(as we discussed yesterday with you: #641),
i managed to do it for now but if you change the arrangement it will be a mess on my end

P.S. If it's too much trouble maybe just something to indicate the end will be sufficient also.

funilrys:
This patch fix #644.

# Warning before merging

As there was too much different information between #644 and https://github.com/StevenBlack/hosts/issues/634#issuecomment-394546359. I take this occasion for clarification.

So before merging we may have to be clear about the marker to use as:

Steven @StevenBlack suggested:
> Asking because, for parsing purposes, I know that `# start {source name}` and # end `{source name}` seems easier. 

@dnmTX suggested:
>```
> #  KADhosts Start
> domain.com
> domain.com
> domain.com
> # KADhosts End
>```

And I did: 

```
# Start StevenBlack 
data
# End StevenBlack

# Start UncheckyAds
data
# End UncheckyAds
```

Cheers,
Nissar

### DIFF
--- a/updateHostsFile.py
+++ b/updateHostsFile.py
@@ -632,8 +632,12 @@ def create_initial_file():
     # spin the sources for the base file
     for source in recursive_glob(settings["datapath"],
                                  settings["hostfilename"]):
+
+        start = "# Start {}\n".format(os.path.basename(os.path.dirname(source)))
+        end = "# End {}\n".format(os.path.basename(os.path.dirname(source)))
+
         with open(source, "r") as curFile:
-            write_data(merge_file, curFile.read())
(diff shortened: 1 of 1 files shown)


StevenBlack:
Thank you Nissar @funilrys!

PR #670, opened 2018-06-18
Fix default encoding

nachoaIvarez:
On macOS High Sierra, running built-in python 2.7.10, the script `updateHostsFile.py` throws:

image

This is happening because the program doesn't ensure UTF8 as default, although in some places it's done ad-hoc in the shape of `.decode("UTF-8")` calls.

This PR ensures the default encoding is set to UTF8, so it's less likely the program might fail on different environments.

After this change:
image

Hope this helps.

### DIFF
--- a/updateHostsFile.py
+++ b/updateHostsFile.py
@@ -22,6 +22,7 @@ import sys
 import tempfile
 import time
 from glob import glob
+from imp import reload
 
 import lxml  # noqa: F401
 from bs4 import BeautifulSoup
--- a/updateHostsFile.py
+++ b/updateHostsFile.py
@@ -23,12 +23,12 @@ import tempfile
 import time
 from glob import glob
 
+import lxml  # noqa: F401
+from bs4 import BeautifulSoup
+
 # Set default encoding to UTF8
 reload(sys)
 sys.setdefaultencoding('utf8')
-
-import lxml  # noqa: F401
-from bs4 import BeautifulSoup
 
 # Detecting Python 3 for version-dependent implementations
 PY3 = sys.version_info >= (3, 0)
--- a/updateHostsFile.py
+++ b/updateHostsFile.py
@@ -22,24 +22,24 @@ import sys
 import tempfile
 import time
 from glob import glob
-from imp import reload
 
 import lxml  # noqa: F401
 from bs4 import BeautifulSoup
-
-# Set default encoding to UTF8
-reload(sys)
-sys.setdefaultencoding('utf8')
 
 # Detecting Python 3 for version-dependent implementations
 PY3 = sys.version_info >= (3, 0)
 
 if PY3:
     from urllib.request import urlopen
+    from importlib import reload
     raw_input = input
 else:  # Python 2
     from urllib2 import urlopen
     raw_input = raw_input  # noqa
+
+# Set default encoding to UTF8
+reload(sys)
+sys.setdefaultencoding('utf8')
 
 # Syntactic sugar for "sudo" command in UNIX / Linux
 if platform.system() == "OpenBSD":
--- a/updateHostsFile.py
+++ b/updateHostsFile.py
@@ -23,6 +23,10 @@ import tempfile
 import time
 from glob import glob
 
+# Set default encoding to UTF8
+reload(sys)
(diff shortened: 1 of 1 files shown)


StevenBlack:
Hi Nacho @nachoaIvarez!  Thank you for this.

As you can see, Travis is complaining.  It looks like our Python linter isn't liking some aspect of this.

Can you please amend this PR with a fix for that?  I'm guessing the linter is not liking code mixed-in with the `import` declarations.  That's the only thing I see, otherwise, looks great!

funilrys:
This PR is great but what the community think about https://stackoverflow.com/a/17628350/5308363 ?

> This is not a safe thing to do, though: this is obviously a hack, since sys.setdefaultencoding() is purposely removed from sys when Python starts. Reenabling it and changing the default encoding can break code that relies on ASCII being the default (this code can be third-party, which would generally make fixing it impossible or dangerous).

funilrys:
@nachoaIvarez normally such issue can be fixed with the following (what I'm doing when working with a terminal which is restricted to ASCII data.

# Linux
```
export LC_ALL=C.UTF-8
export LANG=C.UTF-8
```

# Most other UNIX systems

(Please replace with your country "code")
```
export LC_ALL=en_US.utf-8
export LANG=en_US.utf-8
```

# Start the script

```
python updateHostsFile.py --auto
```

StevenBlack:
/remind me to close this PR in 3 days if no reply from @nachoaIvarez

StevenBlack:
@bkeepers sorry to bug you.  I have @probot reminders configured.  The [docs](https://github.com/probot/reminders/blob/master/README.md) seem to indicate we should see an ack here?

nachoaIvarez:
Sorry guys for the late. The World Cup is really having my complete attention 😅. So, what do you want the outcome of all this to be? Should I add a couple of lines to the docs to account for the requirement of `LANG` and `LC_ALL` variables? Change the code so the linter doesn't fails?

Regarding the quote from @funilrys, I want to remember that the code _already_ counts on UTF-8 to be the encoding, given all the `.decode('UTF-8')` calls in the existing code. So, either we remove all of them, or we set the default encoding at top to make sure all the existing code works.

I think working-by-default is what we should aim for here, but this is just my opinion.

BTW, great work @StevenBlack, I love your project and I'm grateful for your investment in it.

StevenBlack:
I agree with Nacho @nachoaIvarez when he says, we want this working by default.

Guys (and gals), I have a question for readers outside North America and Western Europe.

If we drop support for Python 2.7, to what degree does all this become easier?

Deprecating Python 2.7 would certainly shrink and simplify the code in some places...

funilrys:
Well, we then can merge this. Too bad nobody really answered my question but will get my answer by myself elsewhere ...

In my opinion, dropping the support for 2.7 have to be done today or in >= 2 years (cf. https://pythonclock.org/). So it's just a matter of time.


I will just point out: https://twitter.com/llanga/status/1009104547866742786 (Take the time to understand both side and you may understand the problem of Python 2 vs Python 3 and Experienced vs Inexperienced users ...)

Cheers

StevenBlack:
Nissar @funilrys, not merging this with failing lint under Travis.   Ping Nacho @nachoaIvarez!

nachoaIvarez:
Great! Submitting the updated PR with linter compliance now.

nachoaIvarez:
Added a new commit to include the `reload` import. While in Python 2.x, this was a builtin, but in 3.x, it's in the importlib module.

nachoaIvarez:
Travis was still failing on python 3, because the `imp` module was deprecated in Python 3.4 in favor of the `importlib` module. Added a new commit to do the proper import for python 3, specifically.

StevenBlack:
Thanks Nacho @nachoaIvarez for tending to this.

Oh the joys of maintaining codebases across incompatible versions.

I think we'll drop Python 2.7 support sooner than later.

Maybe we abandon this path, and recommend what Nissar @funilrys suggests in [this comment above](https://github.com/StevenBlack/hosts/pull/670#issuecomment-398176613)?

StevenBlack:
Closing.

PR #671, opened 2018-06-18
update readme to include instructions for clearing dns cache on Mint

mebigfatguy:
Linux Mint is in the tree of Debian/Ubuntu and so you'd expect that those instructions would work. However Mint has its own configuration. As Mint is one of the most popular distros, probably worth having.

### DIFF
--- a/readme_template.md
+++ b/readme_template.md
@@ -304,6 +304,8 @@ Open a Terminal and run with root privileges:
 
 **Debian/Ubuntu** `sudo /etc/rc.d/init.d/nscd restart`
 
+**Linux Mint** `sudo /etc/init.d/dns-clean start`
+
 **Linux with systemd**: `sudo systemctl restart network.service`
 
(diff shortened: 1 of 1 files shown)


StevenBlack:
Thanks Dave @mebigfatguy!  That's a perfect PR, right there.

Except for one thing – god I really should document this but I'm pretty sure it won't help.

`readme.md` is generative.  It's produced from data merged-in at the end of the Python build process.  So this change belongs in [readme_template.md](https://github.com/StevenBlack/hosts/blob/master/readme_template.md).

Dave, can you amend this PR, making the change in `readme_template.md`?  I could do it, but then you won't get credit for being a [contributor](https://github.com/StevenBlack/hosts/graphs/contributors).

mebigfatguy:
no problem, there you go.

StevenBlack:
That's what I'm talkin' about!  Brosius!

PR #677, opened 2018-06-23
Xiaomi tracking and suspicious domains

anudeepND:
Xiaomi uses an app named "msa" which can install apps and updates without notifying the user. Also some of these domains are used by an app named "analytics". These domains doesn't block any legitimate services. #493


anudeepND:
@StevenBlack someone reverse engineered "analytics" app, here's the [link](http://blog.thijsbroenink.com/2016/09/xiaomis-analytics-app-reverse-engineered/)

StevenBlack:
Thanks @anudeepND.  Merging.

PR #680, opened 2018-06-23
Update cws.conviva.com tracking method

xetorixik:
As shown at https://github.com/StevenBlack/hosts/pull/587 cws.conviva.com is already present.
The problem is that the conviva.com tracking method has changed so the blacklist should be changed.

Previously, cws.conviva.com tracked non-generic by cws.conviva.com.
This has changed to unique, specific per user via be43a77aab46b0546578grgrtgr3d4ba6506d488.cws.conviva.com (example).

Various programs like Ad-blockers have the possibility to use wildcards.
In order to keep it so generally working, it seems wise to block the entire domain + sub-domains if the user block software is able to support wildcards.

I tested it with Horizon (local TV service provider), and even though I opted for opt-out of analytics viewing data, it is still being tracked.
Therefore this one should be added immediately.

@ScriptTiger Found [your wild-card entries over here](https://github.com/StevenBlack/hosts/pull/490). 
Could you please take a look at my wild-card commit, whether it is working or not. Thanks.

### DIFF
--- a/data/StevenBlack/hosts
+++ b/data/StevenBlack/hosts
@@ -17,8 +17,6 @@
 127.0.0.1 consumerproductsusa.com
 127.0.0.1 ceromobi.club
 127.0.0.1 com-notice.info
-127.0.0.1 cws.conviva.com
-#*.cws.conviva.com
 127.0.0.1 www.com-notice.info
 127.0.0.1 apple.com-notice.info
 127.0.0.1 www.apple.com-notice.info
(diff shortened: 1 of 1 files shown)


StevenBlack:
Thanks @xetorixik please move this to the `# Spam domains` section.

xetorixik:
Update commit as discussed here #680 (comment) and advised there: #680 (review)

StevenBlack:
Thanks @xetorixik!  Merging.

xetorixik:
Thanks @StevenBlack  , nice one!

PR #723, opened 2018-07-20
New telemetry domains

anudeepND:
@StevenBlack No legit services are afftected by blocking these domains. Let me know if you want more information regarding these domains.
These sites are not present in    
https://raw.githubusercontent.com/StevenBlack/hosts/master/hosts


StevenBlack:
Merging.  Thanks as always, @anudeepND.

PR #726, opened 2018-07-23
Update malwaredomainlist and someonewhocares to use https.

lightswitch05:
Update malwaredomainlist and someonewhocares to use https. This leaves 'MVPS hosts file (http://winhelp2002.mvps.org/)' as the only non-secure source. Interestingly enough, the domain has a valid cert, but all pages results in a 404 - must be some configuration issues.

### DIFF
--- a/data/malwaredomainlist.com/update.json
+++ b/data/malwaredomainlist.com/update.json
@@ -1,9 +1,9 @@
 {
     "name": "Malware Domain List",
     "description": "Malware Domain List is a non-commercial community project.",
-    "homeurl": "http://www.malwaredomainlist.com/",
+    "homeurl": "https://www.malwaredomainlist.com/",
     "frequency": "weekly",
-    "issues": "http://www.malwaredomainlist.com/contact.php",
-    "url": "http://www.malwaredomainlist.com/hostslist/hosts.txt",
+    "issues": "https://www.malwaredomainlist.com/contact.php",
+    "url": "https://www.malwaredomainlist.com/hostslist/hosts.txt",
     "license": "'can be used for free by anyone'"
 }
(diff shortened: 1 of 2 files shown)


lightswitch05:
The build failure is only for 2.7 and looks to be a dependency issue unrelated to this pull request

StevenBlack:
@lightswitch05 I'm considering dropping these Python 2.7 Travis tests...

StevenBlack:
Merging!  Thank you Daniel.

ScriptTiger:
If we have the option to use HTTP to pull from, why wouldn't we? These files don't contain any sensitive data and don't need the added overhead of being encrypted, so HTTP would actually be more efficient in this particular case. I know Windows isn't one to make an example of, but even they use HTTP for Windows Updates. I believe most Linux distros do the same, as well, but I could be wrong there.

Updating the "homeurl" and "issues" to HTTPS is fine because those are optionally opened manually by the user in a Web browser. However, if the script uses the "url" to pull from when it's updating and we end up changing all of our sources to HTTPS, the script will go recursively through one by one likes it's programmed to and actually take longer establishing encryption for each. This may not be noticeable by just changing one or two here or there, but it does add up, especially given the amount of hosts we have.

Am I wrong in how the script operates?

StevenBlack:
@ScriptTiger we've seen situations where `https` fails, for example when certificates expire.

I'm generally happier to run `https` though I agree it's hard to strictly justify.  It protects us from domain spoofing, which is something.

lightswitch05:
@ScriptTiger for someonewhocares.org at least, the https version is significantly faster since the non-secure version just redirects to the secure one. As Steven points out, I think there is something to be said for the integrity aspect of encryption beyond just privacy

PR #730, opened 2018-07-26
Let's say good bye to Python 2!

funilrys:
Good bye Python 2,
thanks for those years of service(s), you will stay forever in our heart !! :tada: :star: :star2: 

------------------

- [x] Announce at the top of the readme.
- [x] Create new branch.
- [x] Remove Python 2 tests.
- [x] Remove Python 2-related code.
- [x] Remove Python 2 mentions in the Readme.

### DIFF
--- a/testUpdateHostsFile.py
+++ b/testUpdateHostsFile.py
@@ -13,9 +13,11 @@ import shutil
 import sys
 import tempfile
 import unittest
+import unittest.mock as mock
+from io import BytesIO, StringIO
 
 import updateHostsFile
-from updateHostsFile import (PY3, Colors, colorize, display_exclusion_options,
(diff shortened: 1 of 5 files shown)


funilrys:
Ouch :sob: wrong branch :sob:

testUpdateHostsFile.py
```diff
@@ -51,10 +46,7 @@ def sep(self):
         return "\\" if sys.platform == "win32" else "/"
```
funilrys: @StevenBlack Why aren't we using https://docs.python.org/2/library/os.html#os.sep there ? :thinking:

StevenBlack: I dunno.  Can you create a new issue to fix this?

PR #736, opened 2018-08-03
Use example files to allow proper git-ignoring

### LINKED ISSUE #144: myhosts and whitelist ignored by .gitignore
The `myhosts` and `whitelist` files seem to have been originally created and checked in to the repo, so git doesn't follow the `.gitignore` file for those two files. Should those be removed from the cache and recommitted, with instructions on how to create those files if you need them? As it exists now, the README is somewhat confusing, since it states that those files aren't tracked but they technically still are.

In addition, should the `hosts` file itself be removed from the cache and added to the `.gitignore` file as well? As it exists now, if I were to add entries to the `myhosts` file, they'll end up getting committed to the base `hosts` file once it's created. You could potentially get burned by that and need to purge your fork. I've removed the `hosts` file from my fork to address that, but that now causes merge conflicts every time I pull.

gfyoung:
The following files:

* `myhosts`
* `blacklist`
* `whitelist`

can be all be modified by the user for personal usage. However, `git` is tracking these files since they exist in the repository, which makes it difficult to do so without accidentally pushing one's own customisations.

This commit converts those examples to ".example" files, which serve as the defaults if one of the files listed above does not exist.

Closes #144.


gfyoung:
@StevenBlack : Test failure is because `flake8` is broken on Python 2.7, which judging from the previous commits to `master`, have been ignored up and until this point.

gfyoung:
@StevenBlack : Friendly ping here.  Any thoughts?

StevenBlack:
@gfyoung I think this is great and I'm definitely going to merge this.  My gut feeling is to merge this and make this active once we've stripped away Python 2.7 support.

Removing python 2.7 support will trigger a major semver bump to version 2.0.0.  

I think this PR goes into the version 2.0.0 milestone, and not here in version 1.4.x.

So may I ask you to keep this PR smelling fresh and ready to merge into the `python3only` branch when times comes?  This PR touches 5 files and now isn't the right time to merge this.  That's my gut feeling.

gfyoung:
@StevenBlack : Certainly!  Would it be possible to add milestones and tag this for `2.0.0`?

gfyoung:
@StevenBlack : How does this look for merging now?

StevenBlack:
@gfyoung a couple of questions.

First, what's the benefit of having identical files `blacklist` and `blacklist.example`, `myhosts` and `myhosts.example`, and `whitelist` and `whitelist.example`?  I don't get why this is materially better?  Refresh my memory.

Second, so far in this PR, we don't mention these `.example` files in the readme?

gfyoung:
The goal was to allow people to properly `.gitignore` their `blacklist`, `myhosts`, and `whitelist` files without accidentally pushing them to GitHub.

The `.example` allows them to do this, and when you execute the script for the first time, the script will auto-generate the `blacklist`, `myhosts`, and `whitelist` files.

StevenBlack:
@gfyoung ah, OK, gotcha.  I hadn't noticed that the `blacklist`, `myhosts`, and `whitelist` files are gone from the base repo and, of course, I checked this after I had tested the generators.

Perfect!  merging.

StevenBlack:
Reminder: since this changes `updateHostsFile.py` and testing, the next release will be a **minor** bump in the version number.

PR #771, opened 2018-09-07
Update docs to say we support Python 3.5+

gfyoung:
Python 3.5 and 3.6 are still alive and well supported.


StevenBlack:
@gfyoung I haven't tested with Python 3.5 or Python 3.6.

But if you say so  😄 

Merging.

PR #772, opened 2018-09-07
Minor house-cleaning

gfyoung:
* No parentheses single-line string
* Remove `builtins` function, which was really for Python 2.x compatibility


StevenBlack:
Thanks @gfyoung good catch!

Merging.

PR #778, opened 2018-09-13
Analytics.ff.avast.com / CCleaner / Avast

xetorixik:
Even though I opted for opt-out of analytics viewing data, usage is still being tracked.
As the "Help improve CCleaner by sending anonymous usage data" is automatically checked again at every restart.

This is confirmed by the firewall logs.
Therefore this one should be added immediately.

No host-block-list contains this analytics entry, hence this commit.

### DIFF
--- a/data/StevenBlack/hosts
+++ b/data/StevenBlack/hosts
@@ -5,6 +5,7 @@
 0.0.0.0 30-day-change.com
 0.0.0.0 2468.go2cloud.org
 0.0.0.0 adservice.google.nl
+0.0.0.0 analytics.ff.avast.com
 0.0.0.0 adsmws.cloudapp.net
 0.0.0.0 androidads23.adcolony.com
 0.0.0.0 analytics.publitas.com


StevenBlack:
Thanks @xetorixik I'm happy to merge this!

xetorixik:
I'm happy to, so lets start the party.
@xetorixik parties all the time!

dnmTX:
If this gets merged they will change the domain in a instant. I've been tracking their domains and there is no end of it.Avast is stilling data more then google and amazon combine(many other domains are involved,the domain pointed by OP is just the beginning). Not saying don't block it but it should be done the smart way to whomever applies(just put it in your hosts file).They will not come to argue here.They have thousands domains ready to go,believe me on this one.

dnmTX:
**Small correction:** even if this gets merged(windows users) it will help with CCleaner but if you using Avast,it has control over your hosts file and it scanning it all the time and everything which contain the word "avast" gets erased,doesn't matter that it could be some malicious domain.They thought about that.

funilrys:
Better avoid Avast and Windows in general then ...

dnmTX:
@funilrys  having all the hosts files in my router solves the Avast/Windows problem.Searching like crazy for new analytics domains because now that it's on the open they'll change it for sure,that's what bothers me.

funilrys:
Better not use Avast nor Windows then ...

dnmTX:
@StevenBlack  new addition: `analytics-v4-poc.ff.avast.com`. It's actually more like a back up(kicks in if the domain from the first post is blocked). Mine as well add it.Let's see how long will take until third back up show up.

PR #779, opened 2018-09-13
Fix of the extra line and the way we manage lines with comments at the end

### LINKED ISSUE #777: Extra newlines in generated hosts file around comments
When I run `python3 updateHostsFile.py`  then scroll to the bottom of the generated `hosts` file I see this:

```
#

# The contents of this file (containing a listing of additional domains in

# 'hosts' file format) are appended to the unified hosts file during the

# update process.



0.0.0.0 someblacklisteddomain.com
```

Which comes from the blacklist file, however in the blacklist file those lines look like this:

```
# blacklist
#
# The contents of this file (containing a listing of additional domains in
# 'hosts' file format) are appended to the unified hosts file during the
# update process.

0.0.0.0 someblacklisteddomain.com
```

Somehow the script is inserting extra newline characters, but only around empty lines (`/^\s?\n/` in regex parlance)  and lines that start with comments  (`/\s?#.*/` in regex parlance). 

Additionally comments after a valid line are being stripped, so for example.

In blacklist...
```
0.0.0.0 someblacklisteddomain.com  # some comment here
```
Becomes...
```
0.0.0.0 someblacklisteddomain.com
```

Interestingly, `myhosts` is not affected by this issue. Inline comments at the end of valid hosts rules are left as-is, and there are no extra newlines between normal comment blocks.

all 3-rd party blocklists and the included custom `blacklist` are both affected by this. Example:

```
# Start adaway.org

# AdAway default blocklist

# Blocking mobile ad providers and some analytics providers

# 
```

funilrys:
Hello there,
This PR fix #777.

This patch:
  * Review of the default state of the `keepdomaincomments` logic.
    * Indeed, comments are now displayed by default.
    * But if you don't need comments, feel free to use the `-k` argument.
  * Delete the required input when calling the `-k` argument.
    * Indeed, before this patch we where obliged to write `-k whatever` but we only check if the string was not empty, so it does not make any sense.
    * Now we only need `-k`.
  * Update tests case regarding the new state of the code.

Have a nice day/night.

Cheers,
Nissar

### DIFF
--- a/testUpdateHostsFile.py
+++ b/testUpdateHostsFile.py
@@ -86,7 +86,7 @@ class TestGetDefaults(Base):
                         "replace": False,
                         "backup": False,
                         "skipstatichosts": False,
-                        "keepdomaincomments": False,
+                        "keepdomaincomments": True,
                         "extensionspath": "foo" + self.sep + "extensions",
                         "extensions": [],
                         "compress": False,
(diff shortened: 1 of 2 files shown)


StevenBlack:
Thanks Nissar @funilrys!

Merging!

I'll update the readme template.

Issue #790, opened 2018-09-27
Could you check https://knowyourmeme.com?


pr10n:
Every time i enter https://knowyourmeme.com my CPU load gets high. Is it just me? I'm thinking about mining or special scripts. Just a suggestion.

dnmTX:
@pr10n  there are 2-3 domains that are not in the lists,not sure how "harmful" they are but you can add them and see if it's better.I personally can't reproduce the issue on my end.
`toolbar.complex.com`
`xhr.acrabakasaka.com`
`assets.pinterest.com`
`api.b2c.com`

pr10n:
I decided to block the knowyourmeme domain itself. I just thought it'd be good to let you know that I have a CPU load issue with that domain just in case it's a reproducible and fixable issue.

ScriptTiger:
Good catch, @dnmTX.
mvps.org hits on stats.complex.com and @lightswitch05 hits on b2c.com, adserver.b2c.com, cdn.adserver.b2c.com, and api-54-190-136-221.b2c.com.

After checking the source, I can only confirm media.complex.com and api.b2c.com as being known domains that should be blocked but aren't, so I'll add those to my current PR. The rest might take some more research.

StevenBlack:
This is fixed in #793.  Thanks everybody!

Closing.

dnmTX:
`xhr.acrabakasaka.com` hosts script that brings ads and tracking once it's loaded.This is taken from the script:

![xhr acrabakasaka com](https://user-images.githubusercontent.com/34774426/46303842-99ad1180-c572-11e8-84bb-6c7a1f2fe7e3.png)

![xhr acrabakasaka com1](https://user-images.githubusercontent.com/34774426/46303847-9f0a5c00-c572-11e8-8c04-77653aa55d88.png)

Ping @lightswitch05

lightswitch05:
@dnmTX thanks, I already have `acrabakasaka.com` and `a.acrabakasaka.com` in the list, happy to add `xhr.acrabakasaka.com` as well

dnmTX:
@lightswitch05  happy to help.Also,can you check `https://civicscience.com/`. Their website is all about surveys,tracking and ads.There is also script from them on that page.

mrjackyliang:
Anyone looking for information on "b2c.com", it belongs to [FouAnalytics](https://fouanalytics.com/). Proof [here](https://securitytrails.com/list/ip/44.233.113.241). Took me a couple of weeks!

[...]

PR #793, opened 2018-09-29
Added some domains for BidGear

ScriptTiger:
The BidGear ad network hasn't been included at all yet and just added some domains for it I just caught.


ScriptTiger:
Just caught another AdTrue domain that slipped by, as well. someonewhocares.org already has "cdn.adtrue.com" on their list, but not "track.adtrue.com"

ScriptTiger:
Added some more caught and discussed in https://github.com/StevenBlack/hosts/issues/790.

StevenBlack:
Tanks @ScriptTiger!  Merging.

PR #836, opened 2018-11-20
Changes to the time.strftime to make it more accurate

spirillen:
Since some host files might need to be updated more frequently it's more appropriate to show timezone and the exact time for last build, hince changed the order of month and date to %d %B

### DIFF
--- a/updateHostsFile.py
+++ b/updateHostsFile.py
@@ -925,7 +925,7 @@ def write_opening_header(final_file, **header_params):
     write_data(final_file, "# This hosts file is a merged collection "
                            "of hosts from reputable sources,\n")
     write_data(final_file, "# with a dash of crowd sourcing via Github\n#\n")
-    write_data(final_file, "# Date: " + time.strftime("%d %B %Y %H:%M:%S (UTC)", time.gmtime()) + "\n")
+    write_data(final_file, "# Date: " + time.strftime("%d %B %Y %H:%M:%S (%Z)", time.gmtime()) + "\n")
 
     if header_params["extensions"]:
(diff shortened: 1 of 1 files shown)


funilrys:
Why don't you use `%Z` instead of the explicit `(UTC)` ?

Otherwise, looks good to me :+1: 

P.S: Why don't we use the [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) format? :smile:

funilrys:
To complete the example of @lightswitch05 we could implement it with:
```python
>> from datetime import datetime, timezone
>> datetime.now(timezone.utc).isoformat(sep=' ', timespec='seconds')
'2018-11-20 22:42:32+00:00'
>> 
```

Docs:
* https://docs.python.org/3/library/datetime.html#datetime.datetime.now
* https://docs.python.org/3/library/datetime.html#datetime.datetime.isoformat

StevenBlack:
@spirillen will you be making the changes suggested by Daniel @lightswitch05 and Nissar @funilrys?

funilrys:
Between, I made a mistake.

It should be:

```python
>> from datetime import datetime, timezone
>> datetime.now(timezone.utc).isoformat(timespec='seconds')
'2018-11-30T05:28:49+00:00'
>> 
```

Indeed, it does not comply to ISO 8601 without the `T`.

spirillen:
Hi Sorry for the long response but have been down with pneumonia

But good suggestions, let me follow up on this

StevenBlack:
Thanks @spirillen.  Merging.

PR #837, opened 2018-11-21
Exclude massive logout sites

borisschapira:
Some sites automatically disconnect users, without their consent, from a number of sites. In addition to being unethical, these practical jokes can lead to the loss of the current contributions on these platforms. I propose to exclude them through a new `logout` extension.

### DIFF
--- a/extensions/logout/update.json
+++ b/extensions/logout/update.json
@@ -1,0 +1,9 @@
+{
+    "name": "Boris' collection of logout jokes domains",
+    "description": "A collection of category-specific host files.",
+    "homeurl": "https://github.com/borisschapira/hostfiles",
+    "frequency": "occasional",
+    "issues": "https://github.com/borisschapira/hostfiles/issues",
+    "url": "https://raw.githubusercontent.com/borisschapira/hostfiles/master/logout-host",
+    "license": "MIT"
+}
--- a/readme_template.md
+++ b/readme_template.md
@@ -57,7 +57,7 @@ Extensions are optional, and are added to the base hosts file.  Extensions are c
 For example, you may want to block porn domains in addition to the adware and malware we block by default.  That hosts file is stored in the porn subfolder of the [`alternates`](https://github.com/StevenBlack/hosts/tree/master/alternates) folder.  
 
 Data for extensions is stored in the [`extensions`](https://github.com/StevenBlack/hosts/tree/master/extensions) folder. You manage extensions by curating the
-[`extensions`](https://github.com/StevenBlack/hosts/tree/master/extensions)  folder tree where you will find the data for `fakenews`, `social`, `gambling`, and `porn` extension data that we maintain and provide for you.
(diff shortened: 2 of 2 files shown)


readme.md
```diff
@@ -88,7 +88,7 @@ Extensions are optional, and are added to the base hosts file.  Extensions are c
 For example, you may want to block porn domains in addition to the adware and malware we block by default.  That hosts file is stored in the porn subfolder of the [`alternates`](https://github.com/StevenBlack/hosts/tree/master/alternates) folder.  
 
 Data for extensions is stored in the [`extensions`](https://github.com/StevenBlack/hosts/tree/master/extensions) folder. You manage extensions by curating the
-[`extensions`](https://github.com/StevenBlack/hosts/tree/master/extensions)  folder tree where you will find the data for `fakenews`, `social`, `gambling`, and `porn` extension data that we maintain and provide for you.
+[`extensions`](https://github.com/StevenBlack/hosts/tree/master/extensions)  folder tree where you will find the data for `fakenews`, `social`, `gambling`, `porn`, and `logout` extension data that we maintain and provide for you.
```
funilrys: It should be in the template file, not here.

extensions/logout/update.json
```diff
@@ -0,0 +1,9 @@
+{
+    "name": "Boris' collection of logout jokes domains",
+    "description": "A collection of category-specific host files.",
+    "homeurl": "https://github.com/borisschapira/hostfiles",
+    "frequency": "occasional",
+    "issues": "https://github.com/borisschapira/hostfiles/issues",
+    "url": "https://raw.githubusercontent.com/borisschapira/hostfilesmaster/logout-hosts",
```
funilrys: I get error 400 when requesting the URL.

dnmTX:
Why extension? why not just merge them in some of the lists here? I never stumbled on any of those sites so far but by just reading about their purpose is enough to know that they should be blocked as they're no better then any of the ad/malware domains here.I'm pretty sure @lightswitch05 would love to have them.

lightswitch05:
Interesting, I’ve never heard of such sites before. Personally I wouldn’t mind them being grouped in to the base list. But they don’t really match my ads and tracking category

borisschapira:
I am open to any suggestion as to whether or not it should be an extension. If you are consensually in agreement to consider this kind of sites as malware, then I'll PR where it is more relevant.

StevenBlack:
Hi Boris @borisschapira!  Thank you very much for this.  

I'm opposed to creating an extension for such a small number of domains.

Currently our raw hosts and variants featuring all combinations of four extensions leads to [16 variants, in total](https://github.com/StevenBlack/hosts/blob/master/makeHosts.py).   That is...

```
1 + C(4, 1) + C(4, 2) + C(4, 3) + C(4,4) = 16
```

Adding one extension leads to **32** variants, that is...

```
1 + C(5, 1) + C(5, 2) + C(5, 3) + C(5,4) + C(5,5) = 32
```

So we won't be doubling the number of variants for such a small number of domains.  🐱

borisschapira:
Roger that. I've opened another PR to merge those domains with your existing ones.

PR #840, opened 2018-11-21
Exclude massive logout sites

borisschapira:
Some sites automatically disconnect users, without their consent, from a number of sites. In addition to being unethical, these practical jokes can lead to the loss of the current contributions on these platforms. I propose to exclude them.


StevenBlack:
Awesome.  Thank you Boris @borisschapira.

Merging.

PR #842, opened 2018-11-23
Added more Google ad domains

ScriptTiger:
Added "tpc.googlesyndication.com" and "googlesyndication.com," caught being used to host YouTube ads.

### DIFF
--- a/data/StevenBlack/hosts
+++ b/data/StevenBlack/hosts
@@ -322,6 +322,8 @@
 0.0.0.0 soperson.com
 0.0.0.0 talk99.cn
 0.0.0.0 www.talk99.cn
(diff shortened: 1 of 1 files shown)


ScriptTiger:
Just as a side note to this, I also noticed YouTube has started pulling ad images from Google image hosting domains like yt3.ggpht.com. Obviously we can't quite block these because they are used for so many other things, but it may be something to keep in mind.

ScriptTiger:
Sorry about this, these domains are already added from other sources, I was cross-checking the wrong list :P

StevenBlack:
Thanks @ScriptTiger – all good!  Keep up the good fight!

PR #849, opened 2018-12-01
Added Ghostery-blocked hosts

JonathanWilbur:
I took this list:

https://raw.githubusercontent.com/bhyde/Ghostery-Scrap/master/report.txt

And transformed it and removed some hosts that are too extreme to block, like "google.com" or "facebook.com".


dnmTX:
This one is too aggressive,seeing for example `aws.amazon.com` in it which will result in too many false positives and i'm pretty sure that there are others that i'm not very much familiar with.
@lightswitch05  you mind take a look,share your opinion.

lightswitch05:
I just took a quick glance but I’m pro duckduckgo.com and godaddy.com seems extream. @dnmTX is right about aws.amazon.com - that’s just generic web hosting that could be anything. Think there might be a lot of duplication with hosts that are already in the list, I noticed a few that I’m pretty sure are already in my list- but I’m on mobile at the moment so I can double check easily

JonathanWilbur:
Yeah, you guys are right. I missed those. And @lightswitch05 , I did deduplicate this list before submitting it. I will close this pull request and re-submit after removing `duckduckgo.com` and `godaddy.com` and `aws.amazon.com`. I will check for others.

dnmTX:
@lightswitch05  here with duplicates removed and converted to UNIX(if you need in DOS let me know):

[no-duplicates.txt](https://github.com/StevenBlack/hosts/files/2636401/no-duplicates.txt)

About 55 duplicates were removed

JonathanWilbur:
No problem! Your judgement makes sense. Thanks for reviewing it anyway.

Issue #760, opened 2018-08-22
Entries break Waze app on Android


timkgh:
```
0.0.0.0 row-advil.waze.com
0.0.0.0 ads-resources.waze.com
0.0.0.0 ads-resources-legacy.waze.com
0.0.0.0 advil.waze.com
```
These entries seem to break the Waze app on Android. The app constantly displays "`Searching Network`" and when trying to access alternate routes it gives an error "`Uh-oh! Failed to communicate with routing server`"

StevenBlack:
Ouch!  Thanks @timkgh

StevenBlack:
Daniel @lightswitch05 can you please check this?  Waze is staggeringly useful...

timkgh:
I should note this does not happen right away when you open the app but after the app is running for a few minutes or some other trigger.
Not sure whether all entries need to be removed or only some of them. Maybe Waze just got smart, if it detects it can't show ads it just claims there's no network.

timkgh:
I will give it a try on a long drive today. Thanks for looking into it.

StevenBlack:
This is fixed in [release 1.4.45](https://github.com/StevenBlack/hosts/releases).

Closing.

timkgh:
Unfortunately the problem is not fixed, I still get the same issue though it takes a lot longer to trigger.
I use `AdAway` with the list `http://sbc.io/hosts/alternates/fakenews-gambling-porn/hosts` and the default 3 lists that come with AdAway.
If I disable just this list, Waze seems to work OK.
Here are the waze related entries in my `/etc/hosts` file:
```
$ grep waze /etc/hosts
127.0.0.1 row-advil.waze.com
::1 row-advil.waze.com
127.0.0.1 ads-resources-legacy.waze.com
::1 ads-resources-legacy.waze.com
127.0.0.1 ads-resources.waze.com
::1 ads-resources.waze.com
127.0.0.1 awazetribune.com
::1 awazetribune.com
```
(last 2 entires just match because of the grep but are not related, some fakenews site it seems)

Not sure whether it's one of the above or something else on the list that may be causing it. Maybe `row-advil.waze.com` needs to be removed too?

If I disable this list but leave the default AdAway lists enabled, all entries above are gone from /etc/hosts.

lightswitch05:
I'll remove row-advil.waze.com later today. Any ideas on how to reproduce would be useful

timkgh:
[...]

PR #859, opened 2018-12-18
added domains serving ads within the Waze navigation app

rajannpatel:


lightswitch05:
There is some history with `advil.waze.com` in #760 - although I was never able to reproduce it personally. Also I think you have a typo- `.cam` vs. `.com`

StevenBlack:
Thanks for the comment, Daniel @lightswitch05.

Rajan @rajannpatel I'm afraid that blocking `advil.waze.com` is more pain than it's worth.

Thank you for the PR, but I must decline it.

Closing.

PR #868, opened 2018-12-29
CoinBlockerList: using main hosts file

FadeMind:
@StevenBlack  IMHO better is using main big hosts file for prevent cryptominers instead of basic browser hosts file (smaller). Works for me.


dnmTX:
I'd say it's about time.I had to remove it from the lists and then load it from the source as an additional hosts file in the past.Good job both of you 👍

PR #876, opened 2019-01-04
grapeshot hosts incorrectly blocked

colinmkeith:
* Removed all grapeshot.co.uk hosts

I work for this company and what we do is to tell an advertiser what a page is about. That is if you own a web page and put a banner ad on it then advertisers want to know as much information about the page and the visitor in order to determine what ad to display and how they are willing to bid to display an ad. We also don't crawl around the web looking for pages, we only request that one of our customers wants to know about.

Most specifically to this list, we don't serve anything to end users from most servers that we run, only to our customers.

The exception to that is a class of customer who embed a HTML script tag directly into their web page to make a call to our servers. In that case they pass the current page's URL and we return the classification directly to the user's browser for the Javascript to use.

In all cases though we return only a snippet of JSON (or JavaScript) that classifies the page. We do not track users, add cookies, serve ads or any of the things that people using block lists don't like. We simply tell an advertiser what a page is about and let them decide what to do with the information.

Most of the entries on this list are wildly wrong. Some are for customers that don't exist any more and won't work and only resolves because it hits a wildcard DNS. E.g.  reed-cw.grapeshot.co.uk.

Others simply don't exist:

% host funke.grapeshot.co.uk
Host funke.grapeshot.co.uk not found: 3(NXDOMAIN)

Others are for the admin tools that our customers use to manage their accounts. E.g.

admin-appnexus.grapeshot.co.uk, admin-thetradedesk.grapeshot.co.uk

Many are simply servers to which we don't see end user traffic. For example:

ss5.ams.grapeshot.co.uk ss2.ash.grapeshot.co.uk ams1.eotws.grapeshot.co.uk

Finally, it seems that someone has taken a personal dislike to our company and added the hostname for our web site to the list, despite the fact that we don't, and never have, served anything but information from this host. It is a regular web site about the company the same way as www.ibm.com talks about IBM.

So again, for the record, what we sell is a service to classify a URL for advertisers to let them know what the site is about so they use that as part of all of the information that they know to decide if they want to bid on an ad on the site. We don't track users, we don't serve ads directly and  we don't collect PII. Most of the hosts listed here are servers that no user will ever connect to, admin sites or a regular web site, none of which are even touched by end users who would make a DNS request and so should not be included in this list.

Thank you.


StevenBlack:
Ping Daniel @lightswitch05

colinmkeith:
Thank you for guiding this into the correct repo and for the changes that you've taken onboard.

PR #877, opened 2019-01-05
Specify encoding to UTF-8 when opening source file

ghostduck:
Hi there, new user here. 

I am just a normal user using this tool to update hosts file. In my Windows machine, I have trouble updating hosts file running `updateHostsWindows.bat`.

```
... (All the Updating sorce data\xxx messages)
Updating source data\malwaredomainlist.com from https://www.malwaredomainlist.com/hostslist/hosts.txt
Updating source data\mvps.org from http://winhelp2002.mvps.org/hosts.txt
Traceback (most recent call last):
  File "updateHostsFile.py", line 1525, in <module>
    main()
  File "updateHostsFile.py", line 164, in main
    update_all_sources(source_data_filename, settings["hostfilename"])
  File "updateHostsFile.py", line 590, in update_all_sources
    update_data = json.load(update_file)
  File "C:\Program Files\Python37\lib\json\__init__.py", line 293, in load
    return loads(fp.read(),
UnicodeDecodeError: 'cp932' codec can't decode byte 0x93 in position 29: illegal multibyte sequence
        1 file(s) copied.
```

I guess this happens since I am using CJKV language options in Windows and Python follows it.

My fix to the issue is specifying encoding to UTF-8. Please see if the 'fix' is acceptable or not.

### DIFF
--- a/updateHostsFile.py
+++ b/updateHostsFile.py
@@ -527,7 +527,7 @@ def update_sources_data(sources_data, **sources_params):
     source_data_filename = sources_params["sourcedatafilename"]
 
     for source in recursive_glob(sources_params["datapath"], source_data_filename):
-        update_file = open(source, "r", encoding="utf8")
+        update_file = open(source, "r", encoding="UTF-8")
         update_data = json.load(update_file)
         sources_data.append(update_data)
         update_file.close()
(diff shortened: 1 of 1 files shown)


ghostduck:
Updated `utf8` to `UTF-8` as requested. 

But I am not really sure about how does it avoid confusion - do you mean you would prefer `git grep -i utf8` would find nothing while `git grep -i UTF-8` find all related codes to encoding?

funilrys:
@ghostduck, let's say that it is a good practice to unify such things as it can be confusing.

I'm not talking about experienced developer (even if it can also happen) but new developer who try to learn or understand with our source code.

I'm not against taking the time to answer questions per issues but with our scale I'm more concerned about escalation as some issues can escalate really fast. And it is really bad the kind of escalation we have here sometimes.

StevenBlack:
Thanks @ghostduck and Nissar @funilrys.

Merging.

PR #919, opened 2019-02-12
IP Logger and URL Shortener Tracking Services

ScriptTiger:
Added domains associated with Grabify IP Logger (https://grabify.link/) and IP Logger URL Shortener (https://iplogger.org/)

These are just the first two services of their kind that pop up in Google search and none of their domains are currently being blocked, except for iplogger.org from someonewhocares.org. Neither require any kind of log-in or account and can both be used completely anonymously. If anyone has time to search around for more of these, please contribute what you can. There are many more such services out there, both with and without log-in/account requirements, and we should work to close this gap as soon as possible.


StevenBlack:

PR #929, opened 2019-03-13
Malware hosts using ttwitter.com not twitter.com

modib:
Not sure which is the right source to fix this so updating hosts.

### DIFF
--- a/hosts
+++ b/hosts
@@ -3444,7 +3444,7 @@ ff02::3 ip6-allhosts
 0.0.0.0 proxy.or4.marketscore.com
 0.0.0.0 proxy.sj3.marketscore.com
 0.0.0.0 proxy.sj4.marketscore.com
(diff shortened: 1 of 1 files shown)


StevenBlack:
Hi @modib, thanks for this.

Our hosts files are generated from sources found in the `data/` subfolders.  It is therefore in the `data/` subfolders where changes need to be made.

Most of the `data/` contents are fetched from other repos and sources on the web.  The companion `update.json` files tell us where the source originates.  If it's an external source, the changes need to be made at source, typically by submitting an issue there.

Closing.

StevenBlack:
@modib the line you suggest changing belongs to Dan Pollock at someonewhocares.org.  Submit this to hosts@someonewhocares.org.

modib:
Thank you!

PR #962, opened 2019-04-25
add a few hostlists

brianclemens:
None of the hostlists I could find seemed to cover my region properly, so I created one.


StevenBlack:
Thanks Brian! @brianclemens.

This is a perfect Pull Request.  Perfect in every way.  I love it.  Brian "gets" it.

Here's what I've decided to do.

Brian, normally I would not accept this PR because we have this rule that our sources must be well curated, which means there needs to be a notable history of list maintenance and updates.

I'm making an exception here because,

* this PR is perfect,
* the data covers Asia, an area we need to expand,
* it includes extensions, which we can do better at growing and maintaining.

**That said** I'm merging this PR but I want you to understand, our rule about curation still holds.   I'd like to see this contribution move forward with updates, several times per year.  In other words, be active and update your repo, whenever you can.

Good?

Presuming we understand each other, merging.  This will show in the next release, coming later today (probably).

brianclemens:
Understood.  I have been using these blocklists on my own routers here for a while now, so maintaining the repo and keeping responsive to issues shouldn't be an issue.  Thanks for merging!

Issue #964, opened 2019-04-28
Process Hacker - Network


Chaython:
In process hacker local address for some of the following processes are reporting the local address as 0.0.0.0
system, steam.exe, svchost.exe, spoolsv.exe, chrome.exe, skypeapp.exe dashost.exe
with my previous hosts file, some of these had a local adress that were either 0koryu0.easter.ne.jp or 24.244.3.32
I read this is because process hacker looks for the first 127.0.0.1 in the hosts file. However in your hosts file https://raw.githubusercontent.com/StevenBlack/hosts/master/hosts 
`127.0.0.1 localhost
127.0.0.1 localhost.localdomain
Local 127.0.0.1
255.255.255.255 broadcasthost
:: 1 localhost
:: 1 ip6-localhost
:: 1 ip6-loopback
fe80 :: 1% lo0 localhost
ff00 :: 0 ip6-localnet
ff00 :: 0 ip6-mcastprefix
ff02 :: 1 ip6-allnodes
ff02 :: 2 ip6-allrouters
ff02 :: 3 ip6-allhosts
0.0.0.0 0.0.0.0
`
are all clearly above the fold, I assume this is an issue with process hacker, however as 0.0.0.0 calls 0.0.0.0 does that mean the applications are calling to a blocked address? I assume then that these are the processes within the apps that are "analytics"? I could just change 0.0.0.0 to localhost in order to hide that these are blocked services?

StevenBlack:
Hi Chaython @Chaython  I'm trying to understand, but I can't seem to make sense of it.

Can you please rewrite that?  I'll try again once you've clarified it.

Chaython:
![image](https://user-images.githubusercontent.com/6486343/57006872-5ca90c00-6ba1-11e9-83ec-fcd28dc784ac.png)
more of an issue with process hacker, though it's pulling local address from hosts file; as you can see some applications have a local address as 0.0.0.0 so is it because these are blocked in the hosts file?
Line 26 refers to 0.0.0.0 0.0.0.0
So if resulting address is 0.0.0.0; the local address is reporting 0.0.0.0
If I change 0.0.0.0 0.0.0.0 to 0.0.0.0 block the local address will be listed as block
So is there any reason you don't call 0.0.0.0 Blockedinhostsfile
![image](https://user-images.githubusercontent.com/6486343/57007204-291bb100-6ba4-11e9-9346-7420c57cec23.png)

dnmTX:
[...]

PR #966, opened 2019-05-01
Update hosts

Chaython:
Changed Line 26's title from 0.0.0.0 to "blocked in hosts file" so novice users using a network monitoring tool such as proccess hacker will see "blocked in hosts file" for applications that traffic is set to 0.0.0.0 rather than seeing "0.0.0.0" and wondering why some applications are reporting that IP rather than localhost

### DIFF
--- a/hosts
+++ b/hosts
@@ -23,7 +23,7 @@ ff00::0 ip6-mcastprefix
 ff02::1 ip6-allnodes
 ff02::2 ip6-allrouters
 ff02::3 ip6-allhosts
-0.0.0.0 0.0.0.0
+0.0.0.0 Blocked_in_hosts_file
 
 # Custom host records are listed here.
 
--- a/updateHostsFile.py
+++ b/updateHostsFile.py
@@ -954,7 +954,7 @@ def write_opening_header(final_file, **header_params):
         write_data(final_file, "ff02::1 ip6-allnodes\n")
         write_data(final_file, "ff02::2 ip6-allrouters\n")
         write_data(final_file, "ff02::3 ip6-allhosts\n")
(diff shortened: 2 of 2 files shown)


Chaython:
I as an end user do not look to build a hosts file; I want the ease of the copy paste from the release; I then the end user may too be confused by a network tool reporting 0.0.0.0 instead of my local machine. 
Thanks for telling me what line to update.

funilrys:
I'm not a big Windows user but you could then write a script similar to the following and run it instead of running the updater as standalone...

**WARNING:** NOT TESTED as I'm under Linux and I'm still trying to learn PowerShell for work :smile_cat:  It's just theoretically working ...

```PowerShell
function Get-Exit-In-Bool($Command) { 
    :: Run command, 
    :: mute output  $?
    :: return true if $Command ran successfully, false otherwise.
    & $Command | Out-Null; $? 
}

if (Get-Exit-In-Bool \PATH\TO\updateHostsWindows.bat) { 
    :: We could run the command successfuly, we replace the line with 0.0.0.0
    ((Get-Content -path %WINDIR%\System32\drivers\etc\hosts -Raw) -replace '0.0.0.0 0.0.0.0','0.0.0.0 WHATEVER') | Set-Content -Path %WINDIR%\System32\drivers\etc\hosts
} else { 
   :: We exit 1, something went wrong
    exit /b 1
}
```

But yeah to quote @dnmTX ([cf](https://github.com/StevenBlack/hosts/issues/964#issuecomment-488231038):
> The lists here is not intended to comply with firewalls or any other programs for that matter.


It's yours Steve @StevenBlack, I'm done with this :smile_cat:

StevenBlack:
Closing.

Issue #982, opened 2019-05-23
Safari attempts to query with "www." prepended causing some blocks to fail


mys721tx:
Some points after Mountain Lion, Safari will attempt to query again with `www.` prepended after the first query failed. ([related question on superuser](https://apple.stackexchange.com/q/92710/37956)) For example, when the following is added:

```
0.0.0.0 example.com
```

Although query to `eample.com` will fail, the query to `www.example.com` does not. Safari is still able to connect to `example.com`. "Show full website address" option is turned on for the screenshots.

```ShellSession
$ dns-sd -G v4 example.com
DATE: ---Thu 23 May 2019---
 2:24:34.094  ...STARTING...
Timestamp     A/R    Flags if Hostname                               Address                                      TTL
 2:24:34.095  Add        2 -1 example.com.                           0.0.0.0                                      1
$ dns-sd -G v4 www.example.com
DATE: ---Thu 23 May 2019---
 2:25:48.835  ...STARTING...
Timestamp     A/R    Flags if Hostname                               Address                                      TTL
 2:25:48.852  Add        2  0 www.example.com.                       93.184.216.34                                3968
```
![image](https://user-images.githubusercontent.com/561975/58230468-64705200-7d02-11e9-9fe0-d1c8ec36de8a.png)
![image](https://user-images.githubusercontent.com/561975/58230489-7d790300-7d02-11e9-9cbf-8210ae94be67.png)

This caused some entries on the block list to fail. Perhaps when generate hosts on macOS, an entry with `www.` should be added to prevent this.

mys721tx:
Additionally, IPv6 may also require blocking. To completely block `example.com`, the following entries are needed in `/etc/hosts`:

```
0.0.0.0 example.com
:: example.com
0.0.0.0 www.example.com
:: www.example.com
```

ScriptTiger:
Personally every time I submit a PR here I always check for WWW domains and make sure they are included, along with any other malicious domains that can be enumerated within reason, and I know several others do the same. Since hosts files are not dynamic and can't use wild cards or REGEX, etc., everything must be explicitly stated.

Now I am not belittling your comments at all, on the contrary I agree that this is indeed a huge issue for those that may be affected by it. However, the solution to this problem would still be PR any WWWs that are leaking through to either the appropriate upstream sources, this being the preferable method, or to the `hosts/data/StevenBlack/hosts` file if for some reason these entries cannot  be appended to upstream sources.

[...]

PR #984, opened 2019-05-24
Adding leaky prefix

mys721tx:
Per #982, a list of domains that have at least one A answers from Google
Public DNS has been added to the list.

I made a tool in Go for this task. ([mys721tx/lpc](https://github.com/mys721tx/lpc)). It still needs some work but it can automate most part of the check.


mys721tx:
Some duplicates were removed, some domains that no longer exist are marked with comments.

StevenBlack:
Hi @mys721tx, thank you for this.

Unfortunately, you've made changes to a generated file.

I think the way this repo works is clear in [the readme](https://github.com/StevenBlack/hosts/blob/master/readme.md).

Our hosts come from a variety of well curated sources, which we amalgamate, de-duplicate, and package in various ways.

The changes you suggest need to happen within curator's repos, not here.

So I must decline this pull request.  But thank you for your efforts!

mys721tx:
The README states that this file is the Steven Black's ad-hoc list, which is not a generated file.

StevenBlack:
My mistake @mys721tx — you're right!  I misread the file path on the [Files changed page](https://github.com/StevenBlack/hosts/pull/984/files).

Reopening.  Thanks for the clarification.

StevenBlack:
Yishen @mys721tx please tell me, what's with the suffix comments?  What do those mean?  How did you establish these, why are they named this way, and why do we want them?

mys721tx:
Those are the errors returned during DNS query. `NXDOMAIN` means the domain in question no longer exist. `SERVFAIL` means that the name server of that domain has failed during query. I am not sure whether those should be removed out right and defer it to you to make that decision.

StevenBlack:
Yishen @mys721tx then I would say, delete the `NXDOMAIN` domain lines, remove the `SERVFAIL` comments, and we're good.

mys721tx:
@StevenBlack removed as requested.

StevenBlack:
Yishen @mys721tx this is all-good.  Thanks!

Merging.

Issue #333, opened 2017-06-03
bad address at /etc/hosts line 5


Ackis:
Noticing this in my error.log:

`Jun  3 15:19:44 ubuntu dnsmasq[28680]: bad address at /etc/hosts line 5`

I'm running dnsmasq 2.75
`Dnsmasq version 2.75  Copyright (c) 2000-2015 Simon Kelley`

```
>lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 16.04.2 LTS
Release:        16.04
Codename:       xenial
```

Here's the first few lines of /etc/hosts:

```
# This hosts file is a merged collection of hosts from reputable sources,
# with a dash of crowd sourcing via Github
#
# Date: June 03 2017
# Number of unique domains: 36,247
#
# Fetch the latest version of this file: https://raw.githubusercontent.com/StevenBlack/hosts/master/hosts
# Project home page: https://github.com/StevenBlack/hosts
#
# ===============================================================

127.0.0.1 localhost
127.0.0.1 localhost.localdomain
127.0.0.1 local
255.255.255.255 broadcasthost
::1 localhost
fe80::1%lo0 localhost
0.0.0.0 0.0.0.0
127.0.1.1 ubuntu
127.0.0.53 ubuntu

# Custom host records are listed here.
```

I update my hosts file via cron:

`0 0     * * *   root            /usr/bin/python /opt/hostupdate/updateHostsFile.py -a -r -f 2>&1 | /usr/bin/logger -t UpdateHosts`

Not sure what else I could provide. Could just be my system, or something more common? Doesn't seem to have any issues though.

StevenBlack:
Hi John @Ackis thanks for this report.

One clarification, please: does the flagged line include the first blank line?

I'm trying to understand if the problem, line 5, is

    255.255.255.255 broadcasthost

or

    ::1 localhost

Ackis:
Good question. It looks like the actual line that's causing the error is:
`#fe80::1%lo0 localhost`

I'm not really sure how I got line 5 - I'm not seeing it in the logs anywhere. Sorry for that misdirection

Played around with the hosts file manually - this gave the error at line 7:

```
# ===============================================================

22.2.2.2 foo.com


127.0.0.1 localhost
127.0.0.1 localhost.localdomain
```

This gave no error:
```
# ===============================================================
127.0.0.1 localhost
127.0.0.1 localhost.localdomain
127.0.0.1 local
[...]

Issue #163, opened 2016-08-17
fe80::1%lo0 localhost error on dnsmasq


braian87b:
this line:

```
fe80::1%lo0 localhost
```

throws error on dnsmasq running on OpenWRT based firmware.

braian87b:
Wed Aug 17 04:04:33 2016 daemon.err dnsmasq[5245]: bad address at /tmp/block.hosts line 6

StevenBlack:
I don't really see what I can do here.  Any suggestions?

braian87b:
this could be fixed when whitelist feature is added, "localhost" could be added.
or maybe just add to Docs a warning saying that fe80::1%lo0 could raise an error when hosts is used on dnsmasq, my dnsmasq didn't resolve and was because of that.

PR #987, opened 2019-05-29
Fix typo in static hosts list

ninadsp:
When `dnsmasq` uses a generated file, it fails with:

`dnsmasq[92249]: bad address at /usr/local/etc/adblock_hosts line 8`

This PR fixes the typo (?) that was causing this error.

### DIFF
--- a/updateHostsFile.py
+++ b/updateHostsFile.py
@@ -948,7 +948,7 @@ def write_opening_header(final_file, **header_params):
         write_data(final_file, "::1 localhost\n")
         write_data(final_file, "::1 ip6-localhost\n")
         write_data(final_file, "::1 ip6-loopback\n")
-        write_data(final_file, "fe80::1%lo0 localhost\n")
+        write_data(final_file, "fe80::1 lo0 localhost\n")
         write_data(final_file, "ff00::0 ip6-localnet\n")
         write_data(final_file, "ff00::0 ip6-mcastprefix\n")
         write_data(final_file, "ff02::1 ip6-allnodes\n")


ninadsp:
https://github.com/StevenBlack/hosts/issues/333 seems like the most relevant issue for this PR. If I understood this correctly, the line should be conditionally inserted in the header. I'm happy to update this PR if such a condition makes sense.

StevenBlack:
Hi Ninad @ninadsp, we've discussed this before.

See for example Issue #333 (***bad address at /etc/hosts line 5***) and, to a lesser extent, Issue #163 (***fe80::1%lo0 localhost error on dnsmasq***).  

So far, we've decided to live with this, because MacOS is so widely used.

But I could be convinced otherwise if this is a showstopper for some.  Is it?

lewisje:
Maybe a better idea would be another parameter or another platform fork, to only add that line if the current platform is MacOS; this would necessitate a content fork, probably one without that line for the main hosts file (for the benefit of unattended router-based deployments) and one with that line for MacOS.

Of course, those router-based deployments *could* just be updated to intelligently detect and strip out that line, and that might be easier than asking the broader and less technical base of MacOS users to make any changes to their own setups.

dnmTX:
Well how about `sed -i '/%lo/d' hosts`  and be done with it,instead of keep on posting and creating more duplicate issues?
It's simple as that for all Linux users.

StevenBlack:
What Dan @dnmTX said.

Closing.

ninadsp:
This works for me, thanks! Was not notified about the conversation, saw the comments now.

PR #1001, opened 2019-06-16
Fixed grammer

Compotation:
Version is a singular verb, so is should be used. The comma before the and is only used when there is a series of 3 or more items.


StevenBlack:
This sentence is absolutely fine.

Closing.

Issue #750, opened 2018-08-10
Reloading host files on Ubuntu


SUlbrichA:
Hi 

Reloading on ubuntu is outdated. Should be 

```
sudo service network-manager restart
```

Thanks for the project!

funilrys:
HI @SUlbrichA and welcome here!

What is your version of Ubuntu ? Why are you talking about `service` when much Ubuntu are now under `systemd`/`systemctl` (since 2014) ?

Cheers,

SUlbrichA:
Hi

the command on the readme did not work under 18.04 so I googled for
refresh/flush DNS and that was on the first hit
Plus, it worked :)

If there is a better alternative I'd be happy to know it (have not tried to
change startup scripts in a looong while)

Cheers
Stef

funilrys:
@SUlbrichA Normally `sudo systemctl restart network.service` should work under Ubuntu as it use systemd :thinking:

StevenBlack:
Thanks everybody.  When I regenerate the next release, it will say 
```
sudo service network-manager restart
```
... for Ubuntu.  This is straightforward and easier to remember, I feel.

Closing.  Thanks Stefan @SUlbrichA!

PR #1002, opened 2019-06-16
Updated Ubuntu/Debian Flush DNS Command to Use Systemd

Compotation:
The command now uses systemctl instead of service.


StevenBlack:
Please see issue #750.

Closing.

PR #1007, opened 2019-06-23
Add some ads sites

Andryshik345:


StevenBlack:
Hi @Andryshik345  – thanks for this input.

Please submit this request to [one of our curated sources](https://github.com/StevenBlack/hosts#sources-of-hosts-data-unified-in-this-variant).

Closing.

Issue #1003, opened 2019-06-19
proposal: release it on pypi


benzkji:
why not release this software as a pypi package? not many things would be needed to do this, and usability for non tech savy / lazy users would greatly go up.

StevenBlack:
Thanks @benzkji.

Does anybody have comments?  Anyone have experience releasing and maintaining packages on [pypi](https://pypi.org/)?   I think this is an interesting idea.

benzkji:
I have very basic experience, from releasing some django packages.

Put together some infos (for setup.py / pypi display), create a manifest that defines files to be copied when installed, make the command(s) available as shell commands, so you dont need to be in the cloned/downloaded project directory to execute them. Would be it, more or less. I've done all of this before, some ages ago ;-)

funilrys:
Hey Steve @StevenBlack!  It's really not a bad idea as we could provide a proper CLI tool. 

I do have a lot of pypi packages 🙂 Will be glad to help with building it along with the automation behind release (with Travis CI for example) 🙃

funilrys:
.

StevenBlack:
Thanks Nissar @funilrys.  Stalebot cuts a wide swath.

Issue #179, opened 2016-09-18
I built the script with Py2exe.


crborga:
I was able to change a few lines and build the script in py2exe if anyone is interested. Being able to replace the hosts within the script would make it complete. Here my modified script.
[HostsFileGenerator.zip](https://github.com/StevenBlack/hosts/files/478783/HostsFileGenerator.zip)

I modified updateHostsWindows.bat as well to run UpdateHostsFile.exe first so it can be run all together.
Also, the telemetry urls are included in myhosts. Just empty that file if you do not want that.

StevenBlack:
Hi @crborga, thank you!

Can you please put this in your own repo, ideally as a new branch of this repo, and submit a [pull request](https://help.github.com/articles/about-pull-requests/)?

Try it! 😄  You'll find this is a much better, more functional way of presenting suggestions and enhancements.

PR #1019, opened 2019-07-20
Support for building exe using pyinstaller

evenprimes:
This is to add very basic support to allow windows users to compile to an exe.

Also updated the batch file to prefer to run the exe, but will still run the .py file if exe isn't found.

### DIFF
--- a/build_exe.bat
+++ b/build_exe.bat
@@ -1,0 +1,2 @@
+pyinstaller --onefile updateHostsFile.py
+copy /y dist\updateHostsFile.exe .
--- a/updateHostsWindows.bat
+++ b/updateHostsWindows.bat
@@ -20,6 +20,11 @@ if not exist "%WINDIR%\System32\drivers\etc\hosts.skel" (
 	COPY %WINDIR%\System32\drivers\etc\hosts %WINDIR%\System32\drivers\etc\hosts.skel
 )
 
+:: Make sure we are in the correct working directory
+rem Path info from:
+rem https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/call#batch-parameters
+cd %~dp0
+
 :: Update hosts file
 if exist updateHostsFile.exe (
     echo run exe
@@ -36,3 +41,4 @@ COPY hosts %WINDIR%\System32\drivers\etc\
 ipconfig /flushdns
 
 :endofscript
+pause 5
--- a/updateHostsWindows.bat
+++ b/updateHostsWindows.bat
@@ -21,10 +21,18 @@ if not exist "%WINDIR%\System32\drivers\etc\hosts.skel" (
 )
 
 :: Update hosts file
-python updateHostsFile.py --auto --extensions fakenews social gambling
+if exist updateHostsFile.exe (
+    echo run exe
+    updateHostsFile --auto --extensions fakenews social gambling
+) else (
+    echo run script
+    python updateHostsFile.py --auto --extensions fakenews social gambling
+)
 
 :: Move new hosts file in-place
 COPY hosts %WINDIR%\System32\drivers\etc\
 
 :: Flush the DNS cache
 ipconfig /flushdns
+
+:endofscript
--- a/updateHostsWindows.bat
+++ b/updateHostsWindows.bat
@@ -21,7 +21,7 @@ if not exist "%WINDIR%\System32\drivers\etc\hosts.skel" (
 )
 
 :: Update hosts file
-pipenv run updateHostsFile.py -a
+python updateHostsFile.py --auto --extensions fakenews social gambling
 
 :: Move new hosts file in-place
 COPY hosts %WINDIR%\System32\drivers\etc\
--- a/updateHostsWindows.bat
+++ b/updateHostsWindows.bat
@@ -21,7 +21,7 @@ if not exist "%WINDIR%\System32\drivers\etc\hosts.skel" (
 )
 
 :: Update hosts file
-python updateHostsFile.py -a
+pipenv run updateHostsFile.py -a
 
 :: Move new hosts file in-place
 COPY hosts %WINDIR%\System32\drivers\etc\
--- a/Pipfile
+++ b/Pipfile
@@ -1,0 +1,13 @@
+[[source]]
+name = "pypi"
+url = "https://pypi.org/simple"
+verify_ssl = true
+
+[dev-packages]
+
+[packages]
+lxml = ">=4.2.4"
+beautifulsoup4 = "==4.6.1"
+
+[requires]
+python_version = "3.7"
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,0 +1,60 @@
+{
+    "_meta": {
+        "hash": {
+            "sha256": "d1158acd995f82e16e190358f87600a63791601ffa925b658b72e45b2f3759c6"
+        },
+        "pipfile-spec": 6,
+        "requires": {
+            "python_version": "3.7"
+        },
+        "sources": [
+            {
+                "name": "pypi",
+                "url": "https://pypi.org/simple",
+                "verify_ssl": true
+            }
+        ]
+    },
+    "default": {
+        "beautifulsoup4": {
+            "hashes": [
+                "sha256:272081ad78c5495ba67083a0e50920163701fa6fe67fbb5eefeb21b5dd88c40b",
+                "sha256:5a3d659840960a4107047b6328d6d4cdaaee69939bf11adc07466a1856c99a80",
+                "sha256:bd43a3b26d2886acd63070c43da821b60dea603eb6d45bab0294aac6129adbfa"
+            ],
+            "index": "pypi",
+            "version": "==4.6.1"
+        },
+        "lxml": {
+            "hashes": [
+                "sha256:06c7616601430aa140a69f97e3116308fffe0848f543b639a5ec2e8920ae72fd",
+                "sha256:177202792f9842374a8077735c69c41a4282183f7851443d2beb8ee310720819",
+                "sha256:19317ad721ceb9e39847d11131903931e2794e447d4751ebb0d9236f1b349ff2",
+                "sha256:36d206e62f3e5dbaafd4ec692b67157e271f5da7fd925fda8515da675eace50d",
+                "sha256:387115b066c797c85f9861a9613abf50046a15aac16759bc92d04f94acfad082",
+                "sha256:3ce1c49d4b4a7bc75fb12acb3a6247bb7a91fe420542e6d671ba9187d12a12c2",
+                "sha256:4d2a5a7d6b0dbb8c37dab66a8ce09a8761409c044017721c21718659fa3365a1",
+                "sha256:58d0a1b33364d1253a88d18df6c0b2676a1746d27c969dc9e32d143a3701dda5",
+                "sha256:62a651c618b846b88fdcae0533ec23f185bb322d6c1845733f3123e8980c1d1b",
+                "sha256:69ff21064e7debc9b1b1e2eee8c2d686d042d4257186d70b338206a80c5bc5ea",
+                "sha256:7060453eba9ba59d821625c6af6a266bd68277dce6577f754d1eb9116c094266",
+                "sha256:7d26b36a9c4bce53b9cfe42e67849ae3c5c23558bc08363e53ffd6d94f4ff4d2",
+                "sha256:83b427ad2bfa0b9705e02a83d8d607d2c2f01889eb138168e462a3a052c42368",
+                "sha256:923d03c84534078386cf50193057aae98fa94cace8ea7580b74754493fda73ad",
+                "sha256:b773715609649a1a180025213f67ffdeb5a4878c784293ada300ee95a1f3257b",
+                "sha256:baff149c174e9108d4a2fee192c496711be85534eab63adb122f93e70aa35431",
+                "sha256:bca9d118b1014b4c2d19319b10a3ebed508ff649396ce1855e1c96528d9b2fa9",
+                "sha256:ce580c28845581535dc6000fc7c35fdadf8bea7ccb57d6321b044508e9ba0685",
+                "sha256:d34923a569e70224d88e6682490e24c842907ba2c948c5fd26185413cbe0cd96",
+                "sha256:dd9f0e531a049d8b35ec5e6c68a37f1ba6ec3a591415e6804cbdf652793d15d7",
+                "sha256:ecb805cbfe9102f3fd3d2ef16dfe5ae9e2d7a7dfbba92f4ff1e16ac9784dbfb0",
+                "sha256:ede9aad2197a0202caff35d417b671f5f91a3631477441076082a17c94edd846",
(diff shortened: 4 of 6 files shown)


StevenBlack:
Thanks Chris @evenprimes!   I'm presently in transit so it may take a few days for me to evaluate this.

Can one of our regular contributors, who is also a Windows user, look this over, please?

funilrys:
My only question, what does this give that is not already possible with "plain" python? 

Cause tomorrow if we implement #1003 we won't ever need this as we may request users to install directly from PyPI, use `python setup.py install` or use `pip` directly from the clone of this repo...

evenprimes:
> 
> 
> My only question, what does this give that is not already possible with "plain" python?
> 
> Cause tomorrow if we implement #1003 we won't ever need this as we may request users to install directly from PyPI, use `python setup.py install` or use `pip` directly from the clone of this repo...

There is nothing that this does that can't be done in plain Python.

The issue I had was that when I had the project installed in a virtual environment (using pipenv) I couldn't get the batch file to correctly run as admin, even with switching the command to `pipenv run updateHostsFile.py <other options>`.

By making the exe I was able to create a run as admin shortcut to the batch file and now it works without messing with the virtual environment.

Note: after a quick glance at [1003](https://github.com/StevenBlack/hosts/issues/1003), I don't see that there is necessarily a conflict. My patch keeps the ability to run the .py file in the batch file, and I put the PyInstaller in a different requirements-exe.txt file so that running as an exe would be 100% optional.

funilrys:
@evenprimes If we follow #1003, a proper CLI tool with a systemwide/user wide callable executable will be created so you'll have to adjust your PR to follow the PATH...

Took the time to read the files ...

May I ask why do you use `pipenv`?

Indeed, you mention the following (also a typo in the word administrator) so why using virtualenv/pipenv?

> Running in a virtual environment in Windows, as admistrator, is a bit of a pain.

I'm wondering now how many people have a problem with that admin permission ... If nobody mentioned anything after so much time (At least I don't remember ready something about an issue with administration run), it may look like this PR is really for your needs.

Steve @StevenBlack will decide but if you run with `pipenv` you may be out of the scope of the recommendation of this repository... So again, back to your scope/your needs...

I'm not against `pipenv` but the fact that you include it (and its deps/structure) into this project where it is only needed for Windows and probably for your needs and a need of few is a bit "huge" for me ... You almost force us to use it and/or maintain it without using it... Which is something user should maintain depends on their needs ... not us as distributer and/or code provider. So for me `pipenv` provides only a solution for your scope. 

What is wrong with that the [README states](https://github.com/StevenBlack/hosts#updating-hosts-file-on-windows)?

> This file MUST be run in command prompt with administrator privileges in the repository directory. In addition to updating the hosts file, it can also replace the existing hosts file, and reload the DNS cache. It goes without saying that in order for this to work, you must be connected to the internet.

Why do I say "force us to use it and/or maintain it without using it" ? 

It's simple you are including `Pipfile` and `Pipefile.lock` which Steve will have to manually push/check/maintain after each update and PR which include new deps into the updater.  And as he, I and probably other contributors of the updater tool don't necessarily use `pipenv`, it will create a situation where at some point the "bare" version is up to date but after 2 years somebody will notice that we totally forgot to update `Pipfile` and `Pipefile.lock` which then create a storm of issue just because none of us use `pipenv` or think about manually updating those files.... 
It's an exaggeration but on our scale, it's something which can happen fast ...

What do the following means?

>I couldn't get the batch file to correctly run as admin?

Can you avoid `readme_exe.md` and add information into `readme_template.md` ? 

If Steve @StevenBlack choose to adopt this PR and your `readme_exe.md`, please add information about the extra files you created (especially `requirements-exe.txt`) and the way to interact with/use them.


Hey, Steve @StevenBlack is this not better as a repo which we can link into the readme? :thinking: 
Or even better have a Windows "guy" (sorry girls I still love you!) who provide a .exe for those who need it? I'm wondering now if @ScriptTiger don't already have something for that purpose :joy_cat:


-----------------------
Again I'm not Steve @StevenBlack ... But that's my opinion.

Cheers

updateHostsWindows.bat
```diff
@@ -20,11 +20,25 @@ if not exist "%WINDIR%\System32\drivers\etc\hosts.skel" (
...
+:: Make sure we are in the correct working directory
+rem Path info from:
+rem https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/call#batch-parameters
+cd %~dp0
+
 :: Update hosts file
-python updateHostsFile.py -a
+if exist updateHostsFile.exe (
+    echo run exe
+    updateHostsFile --auto --extensions fakenews social gambling
```
funilrys: Where does `updateHostsFile` come from ? Is it specific to your environment/scope?

updateHostsWindows.bat
```diff
@@ -20,11 +20,25 @@ if not exist "%WINDIR%\System32\drivers\etc\hosts.skel" (
...
+) else (
+    echo run script
+    python updateHostsFile.py --auto --extensions fakenews social gambling
```
funilrys: You can't force default extensions... It's again your scope...

updateHostsWindows.bat
funilrys: You can't force default extensions... It's again your scope...

updateHostsWindows.bat
```diff
@@ -20,11 +20,25 @@ if not exist "%WINDIR%\System32\drivers\etc\hosts.skel" (
...
+)
 
 :: Move new hosts file in-place
 COPY hosts %WINDIR%\System32\drivers\etc\
 
 :: Flush the DNS cache
 ipconfig /flushdns
+
+:endofscript
```
funilrys: Can you explain this to me? I'm not a windows user but it looks strange that this "appears" from nowhere

updateHostsWindows.bat
```diff
@@ -20,11 +20,25 @@ if not exist "%WINDIR%\System32\drivers\etc\hosts.skel" (
...
+pause 5
```
funilrys: Why a pause?

updateHostsWindows.bat
```diff
@@ -20,11 +20,25 @@ if not exist "%WINDIR%\System32\drivers\etc\hosts.skel" (
 	COPY %WINDIR%\System32\drivers\etc\hosts %WINDIR%\System32\drivers\etc\hosts.skel
 )
 
+:: Make sure we are in the correct working directory
+rem Path info from:
+rem https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/call#batch-parameters
+cd %~dp0
```
funilrys: Why `cd` if we assume that we already are in the correct dirctory?

build_exe.bat
```diff
@@ -0,0 +1,2 @@
+pyinstaller --onefile updateHostsFile.py
+copy /y dist\updateHostsFile.exe .
```
funilrys: Why aren't you moving instead of copy?

funilrys:
@evenprimes Don't take my comment(s) and question(s) too personally 😉 Your PR is awesome and thanks for your time!! 

It's just that I have some opinions about some details...

updateHostsWindows.bat
```diff
@@ -20,11 +20,25 @@ if not exist "%WINDIR%\System32\drivers\etc\hosts.skel" (
...
 
+:: Make sure we are in the correct working directory
+rem Path info from:
+rem https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/call#batch-parameters
+cd %~dp0
+
 :: Update hosts file
-python updateHostsFile.py -a
+if exist updateHostsFile.exe (
+    echo run exe
```
funilrys: Do we really need the echo?

StevenBlack:
Thanks Nissar @funilrys for your input on this.

It will take me awhile to get to this.  Summer weekends in Canada are precious 😎 

As a matter of principle, Chris @evenprimes I totally "get" the Windows run-as-Administrator thing — that's the bane of my existence, recently.  

I can tell you, as an ex-Windows user (many years ago), I don't miss that shitshow.  No offence intended, Windows just isn't my gig anymore, and I couldn't be happier.  My recent experience with run-as-Administrator just reinforces that.

Here's the thing.

Awhile ago we had an issue with Windows users who use notepad.exe for their editor. These poor people were having `CR+LF` issues: host lines were all merged onto one line in the notepad.exe UI. While I sympathize with that, I can tell you, I want no part of it.  

The next billion users can, please, stay well away from this repo.  People who use notepad.exe are doubtless wonderful folks.  I just don't want to deal with that segment.  Because I can easily imagine non-technical Windows users consuming all the bandwidth here, and this will stop being fun.

So I have a natural tendency to say, maybe this should be a separate repo, an extension to this one.

**BUT** let me have a closer look at this, starting Monday probably.  Maybe my outlook will change, based on what I see.  I'm willing to be convinced, but my pre-disposition is not to dilute the focus of this repo.

I hope this sounds, and comes across, as sensible.

ScriptTiger:
Just as @evenprimes references https://github.com/StevenBlack/hosts/issues/179 as the basis for this thread, I have no doubt this thread will reach the same outcome for @StevenBlack in its need for isolation to its own repository. As @funilrys also mentioned, once this project has been confirmed by the community as a valid and safe option, it can simply be linked to in the README.

Windows, as I am sure @StevenBlack  will agree, will always carry with it nuances that require special attention that this repo is just not targeted to focus on. This repo already handles support for the main hosts file aggregator as well as discussions on various topics surrounding everything from the technical details of weeding out specific analytic franchises to even the ethics of blocking or allowing certain categories of websites. We should not forget this repo is not an organizational account and is controlled by @StevenBlack personally, with little delegation shared therein. Those who contribute to this repo do so accepting and appreciating that fact. Adding a Windows beast to the mix would be distracting from the primary focus. Just as this repo prides itself on exclusively including highly curated lists, the components of this repo should also be curated to a high degree, which depends on a high degree of focus. No one that I am aware of is making any money from this community and we all have day jobs, so this is just a simple case of time management.

updateHostsWindows.bat
evenprimes: The extension was in the `.bat` file when I installed.

updateHostsWindows.bat
evenprimes: `:endofscript` is a batch file label and I added it as part of my testing/development. It allowed me to put `goto endofscript` after the part I was testing without having to comment out or mess with the rest of the lines.

updateHostsWindows.bat
evenprimes: When running the batch file by clicking on it, as opposed to the command line, it closed right away and I wanted to confirm it was actually doing what I thought it was.

updateHostsWindows.bat
evenprimes: Because it's not a valid assumption. At least on Windows 10, whatever build I'm on, batch files run as administrator have a working directory of `%windir%` and they seem ignore the "Start in" option in Properties. (Yeah, this caused more than a few bad words to figure out.)

build_exe.bat
evenprimes: Um... that's a good question and I don't have a good answer. I think I'm just more used to typing copy.

updateHostsWindows.bat
evenprimes: In this case `updateHostFile` is `updateHostsFile.exe` and the extension just isn't required.

updateHostsWindows.bat
evenprimes: Nope, I just hadn't used the `if` syntax in a batch file before and wanted to test both branches without a full run.

evenprimes: Oh, and nothing personal. :) Everyone is just trying to make the best tools we can.

updateHostsWindows.bat
funilrys: @evenprimes :+1:

updateHostsWindows.bat
funilrys: @evenprimes No please look at the original file present into the master branch!

it was 
```
python updateHostsFile.py -a
```

even your changes describe what was there before ...

updateHostsWindows.bat
funilrys: So not needed in our scale.

updateHostsWindows.bat
funilrys: Okay we can keep it!

updateHostsWindows.bat
funilrys: Strange :thinking: Someone will need to confirm this as we never needed that before...

build_exe.bat
funilrys: Don't create duplicate like that, please...

updateHostsWindows.bat
funilrys: > Nope, I just hadn't used the `if` syntax in a batch file before and wanted to test both branches without a full run.

Great, you can get rid of it then :wink:

funilrys: > Oh, and nothing personal. :) Everyone is just trying to make the best tools we can.

:+1:

evenprimes:
Going to respond to a bunch of stuff here, so finger's crossed I get it right! :)

@SteveBlack re using windows
I'm actually just coming back to Windows after 15 years on the Mac. I needed a new computer and I wanted something portable. iPads (even the "Pro" models) don't offer the programming and tool options I wanted. I would *prefer* MacOS, but I'll just say "butterfly keyboard" and leave it at that.

I did consider Linux, since I'm sure the desktop support has gotten much better since 2000-ish when I last really tried to use it. But I've had to use Windows for work, so it's the devil I know, at least for now.

## pipenv
This was a goof on my part. I didn't mean to include/force pipenv into the project. I'm just trying it as my virtual environment manager right now. 

Also, possibly related, is that pipenv is *supposed* to support a command `pipenv run` which will run the command in the proper virtual environment. I didn't dig too far into this yet, but this doesn't seem to work, which is a part of why I went the exe route. 

Python packaging has always been a little dicey (no matter the platform) so developing in a venv and running as an exe on Windows (or .app on MacOS) has always been my goto solution for scripts requiring non-standard library modules.

## README comments
@funilrys was concerned that the Windows instructions in the project readme were incorrect. They are not, as far as they go. However, those are *manual* instructions and my goal was to setup something I could automate. I think this distinction is a *significant* part of any misunderstanding on why I'm doing this. :)

I want to:
- have `hosts` in a virtual environment 
- run it as admin on Windows
- automate that process so it happens without my intervention

Also, I avoided editing the main README because I didn't want to confuse folks. I thought putting in a separate  file would make it easier to distinguish it as totally optional. Very possibly I was wrong. ;)

## Separate repo?
@ScriptTiger suggested that maybe this belongs as a separate repo. If so, I'll be happy to try to coordinate with @StevenBlack on that. The good news is that the "meat" of my changes are not to the core functionality, so merging changes from here should continue to be trivial.

updateHostsWindows.bat
evenprimes: Ah, sorry, didn't understand the comment!

updateHostsWindows.bat
evenprimes: It's been a long time since I was on Windows before, but it's not the behavior I remember either. That said, this whole "Run as administrator" is also somewhat new to me.

build_exe.bat
evenprimes: Gotcha!

llacb47:
Curse you stalebot ;)

PR #1029, opened 2019-08-14
Adding updatefaker.com

borisschapira:
Prank site that makes you believe your computer is updating

### DIFF
--- a/data/StevenBlack/hosts
+++ b/data/StevenBlack/hosts
@@ -1154,6 +1154,7 @@
 0.0.0.0 www.myishop.co.uk
 0.0.0.0 www2.myishop.co.uk
 0.0.0.0 ads30.adcolony.com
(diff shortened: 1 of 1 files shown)


StevenBlack:
Merci Boris @borisschapira.

Je ne suis pas convaincu que ce site est malicieux.

@everyone: does anybody feel updatefaker.com is malicious?

FadeMind:
It is not dangerous. It's funny. It's like blocking the MAZE game because you feel shock.

borisschapira:
I had in mind rather to avoid panic among the people targeted by these kinds of jokes, but I understand your reluctance.

spirillen:
I agree with @FadeMind on this :)  Just happy they haven't learned about *nix :yum:

StevenBlack:
Thanks, everybody.  Merci Boris @borisschapira.

PR #1141, opened 2020-02-04
adds try catch and a avoid OSX conflict errors

DonRichards:
```python
# https://github.com/StevenBlack/hosts/issues/590 ModuleNotFoundError for OSX
# sys.path.append('/usr/local/lib/python3.7/site-packages')

try:
    from bs4 import BeautifulSoup
except ImportError:
    from BeautifulSoup import BeautifulSoup

```

Add a in-code correction for site-packages calls for bs4/BeautifulSoup

`sys.path.append` is needed in operating systems running multiple versions of python (I think).

In OSX you'll notice that when you run `$ python -m site --user-site` you can get the user specific site-packages location and append that to the script to correct the issue. Left commented out so that it is only applied by those that need it. Wasn't sure if it should be enabled by default.

Also wrapped the import in a try/catch. This is usually good practice and helps improve the error output. Admittedly this doesn't add a lot of value for this error but it did help me a bit at diagnosing the issue. 
### Before
Screen Shot 2020-02-04 at 8 16 32 AM
### After
Screen Shot 2020-02-04 at 8 16 51 AM

Welcome to close this if you don't feel like this adds value to the repo.

### DIFF
--- a/updateHostsFile.py
+++ b/updateHostsFile.py
@@ -22,7 +22,14 @@ import time
 from glob import glob
 
 import lxml  # noqa: F401
-from bs4 import BeautifulSoup
+
+# https://github.com/StevenBlack/hosts/issues/590 ModuleNotFoundError for OSX
+# sys.path.append('/usr/local/lib/python3.7/site-packages')
+
+try:
+    from bs4 import BeautifulSoup
+except ImportError:
+    from BeautifulSoup import BeautifulSoup
 
 # Detecting Python 3 for version-dependent implementations
 PY3 = sys.version_info >= (3, 0)


funilrys:
Hi @DonRichards  and thanks for your time!

More context is needed.

If you install the dependencies as described in the README, there should be no problem. Indeed, Steven @StevenBlack uses a Mac (Am I wrong ?), so we know that if done right, it is working. 

I don't really see what this PR adds. 


# Side note

If you look at the `requirements.txt`, we requires `beautifulsoup4` ==> https://pypi.org/project/beautifulsoup4/

And not `BeautifulSoup` ==> https://pypi.org/project/BeautifulSoup/.

StevenBlack:
Thanks Don @DonRichards and yes Nissar @funilrys I use Macs, mostly.

Can I close this, Don?

PR #1156, opened 2020-02-22
Fix GitHub capitalization

XhmikosR:


alternates/fakenews-gambling-porn-social/hosts
```diff
@@ -2354,7 +2354,7 @@ ff02::3 ip6-allhosts
 # Blocked: 1,046 domains
 # Only include advertisers in Vietnam
 # Home: https://bigdargon.github.io/hostsVN/
-# Github: https://github.com/bigdargon/hostsVN/
+# GitHub: https://github.com/bigdargon/hostsVN/
```
funilrys: This repository do not change the comments from upstream. That means, that this line stay as it is. 

If you really want to change that, please report to upstream:  https://github.com/bigdargon/hostsVN/

XhmikosR: The repository hasn't changed, it's the text before that that comes from the template that changed.

funilrys: @XhmikosR That line does not come from our template.

XhmikosR:
I just need a clear answer from @StevenBlack or whoever knows, which files are generated and which files they use as templates so that I reduce the patch to only the minimum changes, since the generated files will be updated automatically.

alternates/fakenews-gambling-porn-social/hosts
XhmikosR: OK, I will revert this one. See my comment below. Because I'm not sure what comes from where, I initially replaced every instance.

funilrys:
I'm one of those who know @XhmikosR, 
I'm also one of those who fixes the issue of the updater :smile_cat: 

* All `readme.*` comes from https://github.com/StevenBlack/hosts/blob/master/readme_template.md 
* For the hosts file, what we generate is comes from https://github.com/StevenBlack/hosts/blob/98f6421f444a30ebd260d9a283d612ea270a4cfa/updateHostsFile.py#L996-L1119

    The rest comes from upstream. That means that my comments:
    * https://github.com/StevenBlack/hosts/pull/1156#discussion_r382914807
    *  https://github.com/StevenBlack/hosts/pull/1156#pullrequestreview-363030323
    
    are the only one to fix :-) 

    We can't change the rest as I stated there:
    * https://github.com/StevenBlack/hosts/pull/1156#discussion_r382914807
    * https://github.com/StevenBlack/hosts/pull/1156#discussion_r382915058

I hope that it clarifies a bit the situation.

Cheers!

XhmikosR:
Thanks for the reply! Please check out the last push. I submitted 2 upstream PRs for hostsVN.

StevenBlack:
This is Excellent.  Thanks Nissar @funilrys and thank you @XhmikosR.

Merging.

PR #1157, opened 2020-02-22
Enforce LF

XhmikosR:
This fixes issues on Windows where autocrlf is true by default

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -6,6 +6,9 @@ trim_trailing_whitespace = true
 indent_size = 4
 end_of_line = LF
 
+[*.bat]
+end_of_line = CRLF
+
 [*.md]
 trim_trailing_whitespace = false
 insert_final_newline = true
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,4 @@
+# Enforce Unix newlines for all files
+* text=auto eol=lf
+# Enforce Windows newlines for .bat files
+*.bat text eol=crlf


funilrys:
Force push is never great for others @XhmikosR ...

XhmikosR:
Well, it's great for me to keep the branch clean.

funilrys:
It's a great way to keep a branch clean. But it may cause some issues on the merge. It does not happen every time. But for a few times, I played with it. It broke the copies of all my teammates. 

It's unnecessary noise. I agree it's great to be able to push --force. But on a bigger scale when we are more than 2 people cloning. It can be a nightmare!

My opinion...

XhmikosR:
>It does not happen every time. But for a few times, I played with it. It broke the copies of all my teammates.

This is true 100%, but hey, it's my fork so I'm the only contributor :P

It's not that hard to deal with rebases in unmerged branches, but requires some extra steps. Alternatively, one can just use the option when merging a PR, which isn't being used here judging by the last PR merge.

funilrys:
It's yours Steve @StevenBlack !

StevenBlack:
Nissar @funilrys, do you know, what's the impact of this for the repo, as a whole?

Last thing I want is enormous diffs each time PRs alternate from Windows authors to non-Windows authors.

It's the Windows OS that, from its earliest days, gets line-endings wrong.

XhmikosR:
This actually just enforces the line endings to be LF like in the editor
config file regardless of the OS.

StevenBlack:
Oh I get it now.  Duh!

Merging.

Thanks everyone!

PR #1158, opened 2020-02-22
Use https for winhelp2002.mvps.org

XhmikosR:


StevenBlack:
This is good!  Thanks @XhmikosR.

Merging.

PR #1160, opened 2020-02-22
package.json: add `private: true`

XhmikosR:
The package isn't published on npm.

I haven't tested this with release-it, but it's common practice to have this set to if the package isn't published.


StevenBlack:
Well... whatever.

Merging.

Thanks @XhmikosR

PR #1162, opened 2020-02-23
updateHostsWindows.bat: minor consistency changes

XhmikosR:
* fix a few typos
* fix indentation
* lowercase commands since on Windows case doesn't matter
* add `/y /v` flags in copy commands
* remove unneeded `pushd` command since we `cd` after that anyway
* add `wscript.exe` when calling the .vbs script

Non-whitespace diff: <https://github.com/StevenBlack/hosts/pull/1162/files?w=1>

### DIFF
--- a/updateHostsWindows.bat
+++ b/updateHostsWindows.bat
@@ -1,61 +1,61 @@
 ::
-:: This script will first create a backup of the original or current hosts
-:: file and save it in a file titled "hosts.skel"
+:: This script will first create a backup of the original or the current hosts
+:: file and save it in a file named "hosts.skel".
 ::
-:: If "hosts.skel" exists, the new hosts file with the customized unified hosts
-:: will be copied to the proper path. Next, the DNS cache will be refreshed.
+:: If the "hosts.skel" file exists, the new hosts file with the customized unified
+:: hosts will be copied to the proper path. Next, the DNS cache will be refreshed.
 ::
 :: THIS BAT FILE MUST BE LAUNCHED WITH ADMINISTRATOR PRIVILEGES
 :: Admin privileges script based on https://stackoverflow.com/a/10052222
 ::
-::
-@ECHO OFF
-TITLE Update Hosts
 
-:: Check if we are administrator. If not, exit immediately.
+@echo off
+title Update Hosts
+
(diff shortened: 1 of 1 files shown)


updateHostsWindows.bat
```diff
@@ -1,61 +1,61 @@
...
+del "%TEMP%\getadmin.vbs"
+exit /b
+
+:gotAdmin
+cd /d "%~dp0"
+
+:BackupHosts
+:: Backup the default hosts file
+if not exist "%WINDIR%\System32\drivers\etc\hosts.skel" (
+    copy /y /v "%WINDIR%\System32\drivers\etc\hosts" "%WINDIR%\System32\drivers\etc\hosts.skel"
```
funilrys: The file does not exist. So `/y` is not needed.

XhmikosR: We could probably just rename the file instead of copying it, then.

funilrys: If something went wrong with the updater, you will then break the hosts. That also means that you lose protection for at the runtime of the script.

XhmikosR: `ren` would fail if the file exists anyway. Not important either way since I removed the overwrite flag.

funilrys: Uh ... We are probably talking about the same thing but just to clarify a bit my sentence and what I meant.

You stated:
> We could probably just rename the file instead of copying it, then.

My answer:
> If something went wrong with the updater, you will then lose the hosts. That also means that you lose protection for and at the runtime of the script.

What did I mean? Well, as stated in the comment of the script, it's a backup. 

This means that if we rename `hosts` to `hosts.skel`, for the time of the execution of `updateHostsFile.py`, there will be no protection for the end-user.
So, if something went wrong in the execution of `updateHostsFile.py` you're adding an unexpected behavior (and some work too) for the end-users.

XhmikosR:
I think this is ready for one final review. I tested it on my Windows 10 VM and works fine :)

funilrys:
Great, it's yours Steve @StevenBlack !

StevenBlack:
Thank you @XhmikosR.  And thank you Nissar @funilrys for the assist.

Merging.

PR #1164, opened 2020-02-23
Minor shell script tweaks

XhmikosR:
* always double quote variables
* use https for the miniconda URLs

Would be nice to try this without the cache just in case.

### DIFF
--- a/ci/install_conda.sh
+++ b/ci/install_conda.sh
@@ -1,16 +1,16 @@
 #!/bin/bash
 
 if [ -d "$HOME/miniconda3" ] && [ -e "$HOME/miniconda3/bin/conda" ]; then
-    echo "Miniconda install already present from cache: $HOME/miniconda3"
-    rm -rf $HOME/miniconda3/envs/hosts  # Just in case...
+    echo "Miniconda install already present in cache: $HOME/miniconda3"
+    rm -rf "$HOME"/miniconda3/envs/hosts  # Just in case...
 else
     echo "Installing Miniconda..."
-    rm -rf $HOME/miniconda3  # Just in case...
+    rm -rf "$HOME"/miniconda3  # Just in case...
 
     if [ "${TRAVIS_OS_NAME}" == "osx" ]; then
-        wget http://repo.continuum.io/miniconda/Miniconda3-latest-MacOSX-x86_64.sh -O miniconda.sh || exit 1
+        wget https://repo.continuum.io/miniconda/Miniconda3-latest-MacOSX-x86_64.sh -O miniconda.sh || exit 1
     else
-        wget http://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh -O miniconda.sh || exit 1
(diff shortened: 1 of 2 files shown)


ci/install_conda.sh
```diff
@@ -1,16 +1,16 @@
 #!/bin/bash
 
 if [ -d "$HOME/miniconda3" ] && [ -e "$HOME/miniconda3/bin/conda" ]; then
-    echo "Miniconda install already present from cache: $HOME/miniconda3"
-    rm -rf $HOME/miniconda3/envs/hosts  # Just in case...
+    echo "Miniconda install already present in cache: $HOME/miniconda3"
+    rm -rf "$HOME"/miniconda3/envs/hosts  # Just in case...
```
funilrys: The double quote is not necessary for our use case.

ci/setup_conda_env.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 echo "Creating a Python $PYTHON_VERSION environment"
-conda create -n hosts python=$PYTHON_VERSION || exit 1
+conda create -n hosts python="$PYTHON_VERSION" || exit 1
```
funilrys: Not necessary too.

funilrys:
> Would be nice to try this without the cache just in case. 

Well, it would be great but it's not necessary if we just switch to https.

ci/install_conda.sh
XhmikosR: Even if it's not necessary I went with consistency in this patch.

XhmikosR:
> Your fork normally doesn't have any cache files. So you can test from your fork. Just activate the repository in your Travis CI settings.

Sorry, not going to install Travis CI just for this. Ideally we should just switch to GitHub Actions, but that's for another time.

funilrys:
> Sorry, not going to install Travis CI just for this. Ideally we should just switch to GitHub Actions, but that's for another time.

That won't change today! It's just a switch for your fork inside your Travis CI settings... 

Anyway, It's yours Steve @StevenBlack.

XhmikosR:
It's probably a matter of a few minutes for me to submit a PR in the next days switching to GitHub Actions CI, as long as it works and will be accepted by @StevenBlack  :)

funilrys:
The problem is that later we will have to implement an auto push/release in order to reduce the work of Steve @StevenBlack. It might be easier to do from Travis CI. 

In the current state, it's just for testing the source code so it works well!

XhmikosR:
You can do anything with GitHub Actions CI :) Plus your secrets stay in the repo settings which is a huge plus for me.

funilrys:
My secrets in Travis CI stay encrypted in the configuration file ... But we are talking about cosmetics. 

The most important is that others understand it, and it works without having to change a lot for the maintainer itself.

StevenBlack:
I'm good with this one, too.  Thank you @XhmikosR and Nissar @funilrys too.

Merging.

PR #1163, opened 2020-02-23
Update updateReadme.py

XhmikosR:
* remove duplicate key "issues" in dict literal
* remove trailing whitespaces from the output

The first issue was found with https://lgtm.com/projects/g/StevenBlack/hosts/?mode=list (which is owned by GitHub so feel free to add it to the repo :), the other issues are FPs AFAICT, though)

One last issue I have (but it could be I'm on Windows) is that after running the script some of the Unicode characters aren't outputted properly. For example `Christian Martínez` becomes `Christian MartÃ­nez`. Not sure how to fix this.

PR #1165, opened 2020-02-23
Fix link issue under Windows.

funilrys:
We never saw this, as we are not Windows user ourselves but there is
a real issue.

Indeed, as mentionned by @XhmikosR at
https://github.com/StevenBlack/hosts/pull/1163#issuecomment-590057635
the web links are not correctly formated. Indeed, as we use `sep` to
detect the right directory separator, it's also written into the
`readmeData.json` file.

This patch simply, replace the `\` of Windows, with the universal `/` of
the web convention.

### DIFF
--- a/updateReadme.py
+++ b/updateReadme.py
@@ -45,6 +45,9 @@ def main():
                 "Unified hosts **+ " + key.replace("-", " + ") + "**"
             )
 
+        if "\\" in data[key]["location"]:
+            data[key]["location"] = data[key]["location"].replace("\\", "/")
+
         toc_rows += s.substitute(data[key]) + "\n"
 
     row_defaults = {


XhmikosR:
Oops, just saw this. Refs #1166.

BTW isn't there a better way to handle this stuff? For example Node.js has [path.posix](https://nodejs.org/api/path.html#path_path_posix) and Python seems to have https://docs.python.org/3.5/library/pathlib.html

funilrys:
@XhmikosR The problem is that it comes from `updateHostsFile.py` which really depends on the path of the file. The same path is written into `readmeData.json` which is later used by `updateReadme.py`. 

So no in the current context, I don't see a better solution. Let's keep it simple for others too :wink:

StevenBlack:
Perfect.  Thanks everybody.

Merging.

XhmikosR:
@funilrys it seems there's at least one more case with backslashes:

```diff
-# Fetch the latest version of this file: https://raw.githubusercontent.com/StevenBlack/hosts/master/alternates/fakenews-gambling-porn-social/hosts
+# Fetch the latest version of this file: https://raw.githubusercontent.com/StevenBlack/hosts/master/alternates/fakenews-gambling-porn-social\hosts
```

Happens with `python makeHosts.py` on Windows.

Issue #829, opened 2018-11-13
Give preference to [Unified Hosts AutoUpdate]


Konfekt:
Regarding the section
```
## Updating hosts file on Windows
```
there is a single line at the end of it that points towards other solutions after recommending 
```
updateHostsWindows.bat
```
However, [Unified Hosts AutoUpdate](https://github.com/ScriptTiger/Unified-Hosts-AutoUpdate "Unified Hosts AutoUpdate") by @ScriptTiger brought several advantages on my Windows 10 installation, such as

- replacing the system host file by a host file that left DNS lookup working (perhaps by using `CRLF`?),
- offering to add extensions (gamble, social, ...),
- offering to schedule the update, and
- not requiring the installation of a Python distribution and additional packages such as `beautifulsoup4`. 

Therefore, perhaps the `Windows 10` user should be hinted first towards 

[Unified Hosts AutoUpdate](https://github.com/ScriptTiger/Unified-Hosts-AutoUpdate "Unified Hosts AutoUpdate")

and then to 

```
updateHostsWindows.bat
```

Konfekt:
@StevenBlack Won't this be fixed or would such a pull request be welcomed?

StevenBlack:
I don't understand.

We mention both.   Are you quibbling about the order of presentation?  Of course this repo's native assets come first.  I don't see a problem here.

Closing.

Konfekt:
Yes, this issue is about the order of presentation. The suggestion was to mention  [Unified Hosts AutoUpdate](https://github.com/ScriptTiger/Unified-Hosts-AutoUpdate "Unified Hosts AutoUpdate") by @ScriptTiger first because it brought several advantages on my Windows 10 installation, such as

- replacing the system host file by a host file that left DNS lookup working (perhaps by using `CRLF`?),
- offering to add extensions (gamble, social, ...),
- offering to schedule the update, and
- not requiring the installation of a Python distribution and additional packages such as `beautifulsoup4`.

PR #1169, opened 2020-02-24
Draft proposal for README changes

ScriptTiger:
Addresses issue #829, but also addresses other commonly out-of-scope issues that arise.

As the title suggests, this is only a draft proposal and simply offers an initial suggestion as to changes that might better improve the README. I know that perhaps the styling and format doesn't necessarily flow with that which @StevenBlack keeps, but am looking forward to collaboration on this. I feel as though making changes in this direction might help prevent out-of-scope questions in the future and improve response and action times during the issue triage process. If such issues do persist, at least the documentation sets a solid guideline for the swift action that might be taken against such out-of scope issues and further triage time and effort is not wasted needlessly.

Of course, also, as a community we should always be ready to help those that genuinely need assistance, but that can be further discussed after the issue is closed just so it's not taking priority away from other issues actually directly related to the repository itself and not extraneous indirectly related topics.

### DIFF
--- a/readme_template.md
+++ b/readme_template.md
@@ -1,6 +1,8 @@
 ----
-**Take Note!**  This version of the Hosts file generator, and tests, are for Python 3.5+ only.
-
+**Take Note!**
+- This version of the Hosts file generator, and tests, are for Python 3.5+ only.
+- This repository centers around the hosts file aggregation Python script and NOT around the hosts files it produces. Issues regarding the personal management of hosts files, such as hosts file managers, are out of scope. However, as a courtesy, there are some recommended hosts manager options available.
+- With the exception of issues and PRs regarding changes to `hosts/data/StevenBlack/hosts`, all other issues regarding the content of the produced hosts files should be made with the appropriate data source that contributed the content in question. The contact information for all of the data sources can be found in the `hosts/data/` directory.
 ----
 
 ![readme](https://user-images.githubusercontent.com/36028424/40330477-9df2c2e0-5d7f-11e8-8ac8-511d719a5eae.png)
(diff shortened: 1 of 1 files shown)


readme_template.md
```diff
@@ -1,6 +1,8 @@
 ----
-**Take Note!**  This version of the Hosts file generator, and tests, are for Python 3.5+ only.
-
+**Take Note!**
+- This version of the Hosts file generator, and tests, are for Python 3.5+ only.
+- This repository centers around the hosts file aggregation Python script and NOT around the hosts files it produces. Issues regarding the personal management of hosts files, such as hosts file managers, are out of scope. However, as a courtesy, there are some recommended hosts manager options available.
```
StevenBlack: I feel this is too strong, @ScriptTiger.  Of course we care about the hosts files we produce.

I also feel we're more about the *curation* of hosts sources, than the tool (the python code).  For every release, I read the diffs from sources.  That's more valuable than the Python code which, at core, combines and deduplicates hosts.

I also think, people rarely read the readme attentively.  The Readme is too long, already.  I also think our root folder is so crowded, people currently must scroll-scroll merely to get to the top of the readme.

readme_template.md
```diff
@@ -1,6 +1,8 @@
...
+- With the exception of issues and PRs regarding changes to `hosts/data/StevenBlack/hosts`, all other issues regarding the content of the produced hosts files should be made with the appropriate data source that contributed the content in question. The contact information for all of the data sources can be found in the `hosts/data/` directory.
```
StevenBlack: I think this is basically good, but I doubt this will help.  Because this supposes pre-existing knowledge of the repo's folder structure, and its machinations, both of which are unconventional and not obvious from the repo home page.

readme_template.md
```diff
@@ -255,6 +257,8 @@ Gentoo users may find [`sb-hosts`](https://github.com/PF4Public/gentoo-overlay/t
 
 ## Updating hosts file on Windows
 
+(NOTE: If you love the hosts files produced by this repository but are only interested in managing the hosts files you actually use on Windows and not actually utilizing the full functionality of the awesome script that is the core of this repository, please see the below `Third-Party Hosts Managers` section for hosts manager solutions, such as [Unified Hosts AutoUpdate](https://github.com/ScriptTiger/Unified-Hosts-AutoUpdate "Unified Hosts AutoUpdate").)
+
```
StevenBlack: @ScriptTiger this can be reduced to, "See also some [third-party Hosts managers], listed below.

readme_template.md
ScriptTiger: Maybe something along these lines would be better suited for an issue template, but for now just dropped it.

readme_template.md
ScriptTiger: Done. This will be the second note to refer to that section, so any future issues regarding this will be reasonably exhaustive, I feel. That's why I write so much the first time just to give Windows users that little bit of extra help they may need.

StevenBlack:

PR #1171, opened 2020-02-24
Fix backslashes issue under Windows.

funilrys:
This patch fixes https://github.com/StevenBlack/hosts/pull/1165#issuecomment-590111613.

This patch simply, replace the `\` Windows directory separator, with the
universal `/` of the web convention.


StevenBlack:
Good catch Nissar @funilrys.

Merging.

PR #1199, opened 2020-03-30
xhamster2.desi

John889:
xhamster2.desi

### DIFF
--- a/data/StevenBlack/hosts
+++ b/data/StevenBlack/hosts
@@ -1190,3 +1190,7 @@
 # added March 29, 2020
 0.0.0.0 march262020.best
 0.0.0.0 march262020.com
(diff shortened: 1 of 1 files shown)


John889:
I forgot to add this one:
chat-avenue.com

should I make another request?

you mentioned that reporting the sites to the sources are better than reporting them here. But I don't know which source should I report porn sites to. I would appreciate any suggestions.

StevenBlack:
Thanks John @John889.  This is OK, I'll just raise the curator from here.

Ping @sinfonietta.

StevenBlack:
Thanks @sinfonietta.  

John @John889 this will be live in the next release, probably later today.

Closing.

PR #1204, opened 2020-04-01
fixed 404 url

nickspaargaren:


StevenBlack:
Thank you Nick! @nickspaargaren  Good catch.

Merging.

PR #1210, opened 2020-04-06
CI: specify `set -e`

XhmikosR:
This way the scripts will exit if any error occurs.

### DIFF
--- a/ci/install_conda.sh
+++ b/ci/install_conda.sh
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+set -e
+
 if [ -d "$HOME/miniconda3" ] && [ -e "$HOME/miniconda3/bin/conda" ]; then
     echo "Miniconda install already present in cache: $HOME/miniconda3"
     rm -rf "$HOME"/miniconda3/envs/hosts  # Just in case...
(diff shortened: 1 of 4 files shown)


ci/install_conda.sh
```diff
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+set -e
+
 if [ -d "$HOME/miniconda3" ] && [ -e "$HOME/miniconda3/bin/conda" ]; then
     echo "Miniconda install already present in cache: $HOME/miniconda3"
     rm -rf "$HOME"/miniconda3/envs/hosts  # Just in case...
```
XhmikosR, line 7: BTW @StevenBlack I was wondering doesn't this cover some of the lines from here? https://github.com/StevenBlack/hosts/blob/ecd59a8aee2810e73063afc25594e59a85f69b5b/.travis.yml#L10-L14

Maybe it would be better if all the `before_cache` commands were moved to this script?

StevenBlack:
This is good!  Thanks XhmikosR!

Merging.

ci/install_conda.sh
StevenBlack: @XhmikosR I don't follow.  What is "this" when you say, "*...doesn't this cover some of the lines from here?*"

XhmikosR: @StevenBlack I meant the `rm -rf` lines in the `install_conda.sh` script 🙂

StevenBlack: @XhmikosR I still don't understand.

Can you please make a separate issue for this, and explain what you mean cleanly?  Thanks.

XhmikosR: No worries, it does hurt and since it ain't broken, all good :)

PR #1213, opened 2020-04-06
Add takehotprizzes3.live

breucode:
This site distributes ads. I landed on that site directly from a first-page result from google.

The root site says, site under construction. Should I provide a link to an ad? I am not sure if this would be "good" or appropriate.


StevenBlack:
Thank you @breucode.  Merging.

PR #1225, opened 2020-04-13
Bug fix, plus whitelist/blacklist from command line

djnym:
Adding two things because they are small.

1. adds blacklist and whitelist file command line overrides
2. fix for my previous pull-request, where attempting to generate with a non-existent output directory resulted in an exception.  This appears to have been because of the open with "a" which does not appear to work with a non-existent directory.  In this case I just use os.path.isfile/1 as os.path already appears to be used in other areas of the script, to skip the initial attempt at creation, then use "wb" on the known non-existent file as that appears to create the directory.

I can split into two PR's if having these two in a single one is a problem, but they are both so small I figured it's probably okay.  Let me know if it's not.

### DIFF
--- a/updateHostsFile.py
+++ b/updateHostsFile.py
@@ -201,6 +201,20 @@ def main():
         "ignoring non-necessary lines "
         "(empty lines and comments).",
     )
+    parser.add_argument(
+        "--whitelist",
+        "-w",
+        dest="whitelistfile",
+        default="",
+        help="Whitelist file to use while generating hosts files.",
+    )
+    parser.add_argument(
+        "--blacklist",
+        "-x",
+        dest="blacklistfile",
+        default="",
+        help="Blacklist file to use while generating hosts files.",
+    )
 
     global settings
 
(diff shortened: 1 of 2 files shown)


updateHostsFile.py
```diff
@@ -1296,21 +1310,21 @@ def remove_old_hosts_file(old_file_path, backup):
...
-        shutil.copy(old_file_path, backup_file_path)
+            # Make a backup copy, marking the date in which the list was updated
+            shutil.copy(old_file_path, backup_file_path)
 
-    os.remove(old_file_path)
+        os.remove(old_file_path)
 
     # Create new empty hosts file
-    open(old_file_path, "a").close()
+    open(old_file_path, "wb").close()
```
funilrys, line 1327: Why is `wb` so important?

updateHostsFile.py
```diff
@@ -1296,21 +1310,21 @@ def remove_old_hosts_file(old_file_path, backup):
         Whether or not to backup the existing hosts file.
     """
 
-    # Create if already removed, so remove won't raise an error.
-    open(old_file_path, "a").close()
+    # only go through possible backup and removal if the file exists
+    if os.path.isfile(old_file_path):
```
funilrys, line 1314: let's simplify to:
```
if os.path.isfile(old_file_path) and backup:
```

updateHostsFile.py
```diff
@@ -1296,21 +1310,21 @@ def remove_old_hosts_file(old_file_path, backup):
...
 
-    if backup:
-        backup_file_path = old_file_path + "{}".format(
-            time.strftime("%Y-%m-%d-%H-%M-%S")
-        )
+        if backup:
```
funilrys, line 1316: after simplification, this can be deleted and indentation decreased.

updateHostsFile.py
djnym: I could be "w" probably, I didn't write the original code, but from the comments what I think the purpose was was to make sure the file exists before it's removed, then create it.  I think the second create could be removed without issue at which point this just removes the old file it it exists.

updateHostsFile.py
djnym: That changes the logic, because if you look at the code closely, so see the deletion always happens if the file exists, but backup is only done if the backup option is given.  So if you collapse the conditional, it ends up not working as expected.

updateHostsFile.py
djnym: Actually, when I try some other parameters like '--compress' things fail, because the host file is opened directly in at least 3 different places.  I'm going to make a pass to clean that up in a different PR, so will pull this part of this one.

djnym:
Whitelist/Blacklist parts now in https://github.com/StevenBlack/hosts/pull/1227, closing this to refactor in a better way.

PR #1227, opened 2020-04-13
Allow whitelist/blacklist files to be passed via command line

djnym:


StevenBlack:
Hi Anthony @djnym — this appears to break existing functionality.

The semantics of config here work like this (as you know): **defaults** are overridden by **options** and results in **settings**.

Our defaults are, 

```
...
        "blacklistfile": path_join_robust(BASEDIR_PATH, "blacklist"),
        "whitelistfile": path_join_robust(BASEDIR_PATH, "whitelist"),
...
```

So the defaults are files named, literally, `blacklist` and `whitelist`, with no extensions.

In this PR, the new options assign `""` to both these variables.

When I create a hosts file with no options but with a local `blacklist` file, none of the `blacklist` domains end up in the generated file.

So we need some logic here, and a unit test would be nice, too.

djnym:
I defaulted the args to the same values in the defaults structure, so I think that should fix the regression (at least local testing shows that it does).    I don't currently see any testing around args, other than that the defaults work, so that's a bit of a large ask to add testing there.  I might be able to get to it at some point, but no promises.

StevenBlack:
No problem with the tests, Anthony @djnym.

StevenBlack:
Thanks Anthony @djnym this works great!

Merging.

PR #1231, opened 2020-04-18
Fix FadeMind's lists license

XhmikosR:
@StevenBlack It seems the repo is using the [MIT License](https://github.com/FadeMind/hosts.extras/blob/master/COPYING)


StevenBlack:
Thanks @XhmikosR.  Merging.

PR #1236, opened 2020-04-20
Fix default whitelist argument.

adherzog:
Recent changes in #1227 allowed setting the whitelist and blacklist files via command line arguments. That change set the wrong default for the `--whitelist` option.

This PR changes the default whitelist file from 'blacklist' to 'whitelist'.


StevenBlack:
Nice catch Adam @adherzog.  Thank you.  Merging.

Issue #1166, opened 2020-02-23
Windows script issues


XhmikosR:
After #1157 lands, one of the issues I face on Windows will be fixed. These leaves us with two more issues I've noticed so far:

1. Using updateReadme.py results in broken Unicode characters. Example: `Christian Martínez` becomes `Christian MartÃ­nez`
2. The script results in backslashes used in Readme URLs:
    ```diff
    -Unified hosts **+ fakenews** | [Readme](https://github.com/StevenBlack/hosts/blob/master/alternates/fakenews/readme.md) | [link](https://raw.githubusercontent.com/StevenBlack/hosts/master/alternates/fakenews/hosts) | 52,621 | [link](http://sbc.io/hosts/alternates/fakenews/hosts)
    +Unified hosts **+ fakenews** | [Readme](https://github.com/StevenBlack/hosts/blob/master/alternates/fakenews\readme.md) | [link](https://raw.githubusercontent.com/StevenBlack/hosts/master/alternates/fakenews\hosts) | 52,628 | [link](http://sbc.io/hosts/alternates/fakenews\hosts)
    ```

funilrys:
2 ==> #1165 
1 ==> It comes from the readme_template. It's probably because you're not reading in `UTF-8` ?

ScriptTiger:
I got frustrated with this and just started my own repo, but it's good to see some work being done on this again.

XhmikosR:
@funilrys about the second point, there's still at least once we end up with backslashes, see https://github.com/StevenBlack/hosts/pull/1165#issuecomment-590111613

About the UTF issue, I didn't do anything myself, I just ran the scripts.

XhmikosR:
@StevenBlack there's still an issue with backslashes, see my comment above https://github.com/StevenBlack/hosts/issues/1166#issuecomment-590247036

XhmikosR:
BTW what's the exact scripts/command you run @StevenBlack to generate the files? I'm asking because whenever I try it on Windows I get too many changes in each file

StevenBlack:
@XhmikosR I haven't used Windows in at least 10-years.  I'm on MacOS and Ubuntu.

I run [makeHosts.py](https://github.com/StevenBlack/hosts/blob/master/makeHosts.py) which generates all the variants, in turn.

What do you mean, "I get too many changes in each file"?

XhmikosR:
@StevenBlack understandeable, but please re-open the issue until we manage to fix everything; we are so close :)

[...]

PR #1238, opened 2020-04-21
Unification of the sorting of the sources.

funilrys:
Before this patch, there was no proper sorting of the sources.

As @XhmikosR mentioned in StevenBlack/hosts#1166, without this patch,
the output is totally different in Windows. But let's be honest, if it
is like that under Windows, chances are that the same behavior happens
across different OSes or machines around the globe.

Another reason behind this patch is that - desperate the fact that we
certainly trust @StevenBlack - the integrity of the generated files
could not be guarenteed because of the sorting which may be completely
different from an OS to another and a machine to another.

For those reasons, this patch introduces a unification of the sorting of
all sources.

The idea behind this patch is to have @StevenBlack's ad-hoc hosts
file always on top (1st) and the rest of the sources sorted
alphabetically based on the name of the folder inside the `data`
or `extensions` directory.

This will ensure that we get the same result everywhere.

Concretely speaking, I just added the function (`sort_sources`) which
sorts a given list of sources files. And later on, call the new function
everywhere it was necessary. Tests of the newly introduced function
are also included.

Contributors:
  * @ScriptTiger
  * @XhmikosR

Notes:
  * This patch fix (completely ?) ScriptTiger/hosts#1
  * This patch fix https://github.com/StevenBlack/hosts/issues/1166#issuecomment-590511086

### DIFF
--- a/testUpdateHostsFile.py
+++ b/testUpdateHostsFile.py
@@ -41,6 +41,7 @@ from updateHostsFile import (
     query_yes_no,
     recursive_glob,
     remove_old_hosts_file,
+    sort_sources,
     strip_rule,
     supports_color,
     update_all_sources,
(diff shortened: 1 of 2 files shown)


dnmTX:
@funilrys  please make sure after the patch is applied the **Start...** and **End..** on each list remains.
This is very importand for my scripts to work properly. Thank you 👍

funilrys:
@dnmTX Checked, no changes!

StevenBlack:
Thanks everyone.  Merging.

XhmikosR:
Huge thanks to @funilrys! Now we are consistent across all OS'es.

I'll be able to confirm after a new release is made :)

PR #1246, opened 2020-04-24
Restore filename to follow the hosts-%Y-%m-%d-%H-%M-%S format

gprestes:
On 2d696a42d98212310b80f6085172bd6d67eea7ea the '-' separating hosts from timestamp in backup filename was removed. This pull request is to restore the separator.


StevenBlack:
Thank you Guilherme @gprestes!  

Merging.

PR #1250, opened 2020-04-28
Switch to GitHub Actions CI.

XhmikosR:
I have this patch for a while on  my fork, but being I'm not a Python expert I'm not 100% sure it works like before.

The idea is that one does not rely to any external service and IMO GitHub Actions offer so many possibilities.

A preview can be found on my fork: <https://github.com/XhmikosR/hosts/actions/runs/91155575>

PS. This won't fire until it's merged due to security reasons.

Let me know your thoughts and if you have any suggestions.


.github/workflows/ci.yml
```diff
@@ -0,0 +1,47 @@
...
+  CI: true
+
+jobs:
+  test:
+    runs-on: ${{ matrix.os }}
+
+    strategy:
+      fail-fast: false
+      matrix:
+        python: [3.5, 3.6, 3.7]
```
funilrys, line 13: Let's add 3.8 too 😉

.github/workflows/ci.yml
```diff
@@ -0,0 +1,47 @@
...
+
+      - name: Set up Python
+        uses: actions/setup-python@v1
+        with:
+          python-version: ${{ matrix.python }}
+          architecture: 'x64'
+
+      - run: python --version
+
+      - name: Cache dependencies
```
funilrys, line 28: This is actually optional for me. But we can keep it.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,47 @@
...
+        run: |
+          export PATH="$HOME/miniconda3/bin:$PATH"
+          ci/install_conda.sh
+          ci/setup_conda_env.sh
+
+      - name: Run tests
+        run: |
+          export PATH="$HOME/miniconda3/bin:$PATH"
+          ci/test.sh
+          ci/lint.sh
```
funilrys, line 47: Linting (`ci/lint.sh`) should have it's own "jobs"/"stage" (don't know the word yet 😥).

And it should be done before the test run and not after. :smile_cat: I know it was done like that before, but it does not make sense to lint after the tests ...

XhmikosR: If you check the current CI Travis workflow, you'll see it's like that. This PR just tries to stay as close as the current Travis CI as possible.

So, I won't make this change nor the Python 3.8 addition here. Those should be made separately.

XhmikosR:
Did you confirm the coda env works as expected?

funilrys:
@XhmikosR I looked at your jobs before my review and I assumed that it's working since you added the `set -e` into all CI related files.

The most effective way we have to test the behavior is to add a `python --version` just before running the tests (right before [this line](https://github.com/XhmikosR/hosts/blob/444b36078671deaa77ca8fcc5cd7c4922cb402b6/ci/test.sh#L8)).

XhmikosR:
Alright, I made the changes you requested in a separate patch and also added `python --version`.

The latest run is now here: https://github.com/XhmikosR/hosts/actions/runs/91151524

It seems there's a discrepancy between the CI Python version and the one Conda is installing but I guess that's expected? BTW maybe Conda is not needed anymore? Or we should adapt things later for CI.

.github/workflows/ci.yml
XhmikosR: I don't mind removing it and in fact it's something that has confused me in the past because of there lines:

https://github.com/StevenBlack/hosts/blob/master/ci/install_conda.sh#L5-L10

funilrys:
Awesome it's working as expected! 

We can't get rid of conda. Conda will provide us the latest version (at D-0) of the selected python version when the CI engine/ecosystem may take up to some days before providing us the latest version. 

This gives us an advantage: If the code does not work with the latest Python version, we know that we need to take action (mostly code side). 

Just my 2 cents as Steve @StevenBlack will decide!

Anyway, thanks for your time (again)!

Stay safe and healthy.

StevenBlack:
Thanks @XhmikosR this is great!  Thank you for taking the initiative to do this.

Merging 😃

XhmikosR:
@StevenBlack glad you like it!

BTW do not forget to remove Travis CI (or any other unused stuff) from the repository's apps/integrations/webhooks 🙂

StevenBlack:
Thanks for the reminder @XhmikosR.  Done!

XhmikosR:
A further improvement would be to see if you actually want to cache on CI and revisit this code https://github.com/StevenBlack/hosts/blob/master/ci/install_conda.sh#L5-L10

You should know better than I what's needed there :)

PR #1259, opened 2020-05-03
Make sure we treat any readme assets as `utf-8`

XhmikosR:
@StevenBlack this fixes https://github.com/StevenBlack/hosts/issues/1166#issuecomment-622992965

FYI @funilrys

### DIFF
--- a/updateReadme.py
+++ b/updateReadme.py
@@ -26,7 +26,7 @@ def main():
         "${fmtentries} | "
         "[link](http://sbc.io/hosts/${location}hosts)"
     )
-    with open(README_DATA_FILENAME, "r") as f:
+    with open(README_DATA_FILENAME, "r", encoding="utf-8") as f:
         data = json.load(f)
 
     keys = list(data.keys())
@@ -82,7 +82,7 @@ def main():
         with open(
             os.path.join(data[key]["location"], README_FILENAME), "wt", encoding="utf-8"
         ) as out:
-            for line in open(README_TEMPLATE):
+            for line in open(README_TEMPLATE, encoding="utf-8"):
                 line = line.replace(
                     "@GEN_DATE@", time.strftime("%B %d %Y", time.gmtime())
                 )


StevenBlack:
Looks good to me, but I trust Nissar @funilrys more than I trust myself with nuanced things like this 😃

updateReadme.py
```diff
@@ -26,7 +26,7 @@ def main():
         "${fmtentries} | "
         "[link](http://sbc.io/hosts/${location}hosts)"
     )
-    with open(README_DATA_FILENAME, "r") as f:
+    with open(README_DATA_FILENAME, "r", encoding="utf-8") as f:
```
XhmikosR, line 29: This might be currently needed, but I guess it future proofs in case the json file includes any utf-8 characters in the future.

XhmikosR:
I tested it before making the PR and the only diff I get with this patch compared to master is:

```patch
 readme.md | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)

diff --git a/readme.md b/readme.md
index b2d4fca5..c9f602d6 100644
--- a/readme.md
+++ b/readme.md
@@ -21,7 +21,7 @@
 This repository consolidates several reputable `hosts` files, and merges them
 into a unified hosts file with duplicates removed.  A variety of tailored hosts files are provided.
 
-* Last updated: **May 02 2020**.
+* Last updated: **May 03 2020**.
 * Here's the [raw hosts file with base extensions](https://raw.githubusercontent.com/StevenBlack/hosts/master/hosts) containing 55,945 entries.
 * Logo by [@Tobaloidee](https://github.com/Tobaloidee).
 
```

Which means the issue is totally fixed here 🙂

updateReadme.py
funilrys: Indeed, but `json.loads()` should handle that for us :-) But it's great to stay ahead of the future!

updateReadme.py
```diff
@@ -82,7 +82,7 @@ def main():
         with open(
             os.path.join(data[key]["location"], README_FILENAME), "wt", encoding="utf-8"
         ) as out:
-            for line in open(README_TEMPLATE):
+            for line in open(README_TEMPLATE, encoding="utf-8"):
```
funilrys, line 85: That's what happens when we never test under Windows! 

Would have done the same. Good catch @XhmikosR :clap:

StevenBlack:
Thanks everybody!  Merging!

PR #1275, opened 2020-05-10
hosts: malware IoC

scafroglia93:
Signed-off-by: scafroglia93 <scafroglia57@gmail.com>


StevenBlack:
Lorenzo @scafroglia93 where do these domains come from?

scafroglia93:
AlienVault Threat Exchange from security company like trendmicro, fireye, palo alto and much more

llacb47:
Since there are 1000+ domains here maybe it would be better to find a reputable hosts file/domain list to add.

scafroglia93:
All the domains I have entered are selected directly by AT&T through its authoritarian sources.
I worked about 4 hours to be able to insert them with appropriate checks

This is one of the most used lists, entering these domains could increase coverage in terms of protection against malware, APT, ransomware, phishing and other problems

StevenBlack:
You see Lorenzo @scafroglia93, here's the problem.

All our sources here are *curated*.  This means, our sources add and delete domains based on the judgment of active curators.  This is how we distinguish quality sources from all the others.

**The easiest thing** is create an add-only bucket where domains go in, and never come out.  **The easiest thing**  for us is provide a list with a million domains.

We once evaluated a source with 350,00 domains.  Turns out, this was effectively an add-only bucket, and we declined using it.   Eventually it reached 750,000 domains, and then it (recently) disappeared from the Internet.  Gone.  It was a list that, in the end, had no life.  It was just a dead bucket full of dead shit.

So here you say, here are 1,300 domains.  No links, no curation, just a list.  You dump it into my personal list which says, in effect, "*Here you go Steve, YOU curate these 1,300 domains starting now, and forever.*"

Then I get emails.  Then we see issues raised here.  These contacts say, *"I am from 123abc.com, why is our domain in your list?"*.  And I have no answer for that.  Each one of these costs me at lest 10-minutes, and I invariably remove the domain from the list because, I don't know why it's in the list in the first place.

For each release I personally review the `git diff` for all our sources.  This takes a lot of time.  When I review each `git diff`, I rarely know why a particular domain is included, but I can also see domains that are removed.  The curator cares.  The curator is working to make the list as good as it can be.  I trust our curators.

I believe the 1,300 domains you add are legitimate.  What I don't believe is my ability, and my responsibility, to curate 1,300 additional domains into the future.

scafroglia93:
Schermata da 2020-05-11 19-57-18

This are the report made by alienvault, all was referenced but you have a great 
I make a personal list where I collect alienvault reports; in the meantime you can revert those that I have inserted if you don't like them

XhmikosR:
@StevenBlack I think the wise thing would be to revert any additions that landed on master, and look into a separate list, like you said.

StevenBlack:
I've decided to merge this.  It's valuable.  It's fresh.  I can live with this, for now.

I'm simply erring on the side of caution until a better solution emerges.

But I don't want my ad-hoc list to become a place for the whole Internet to dump big lists of domains.  That won't scale.  Heck, that's already broken.

Ultimately what we need is a curator, or two, who focus on the most current reports.  Responsive and edge-aware curators with latency measured in hours, not days.

scafroglia93:
What do you think of moving everything into a list that I manage directly through the upstream ?

StevenBlack:
Lorenzo @scafroglia93 this is what I am looking for, yes.

We need responsive, leading-edge threat detection.  I can't curate that myself.  Too much going on.

PR #1295, opened 2020-05-27
Update updateHostsFile.py

XhmikosR:
Escape backslashes

@StevenBlack @funilrys Untested, but it makes sense 🙂 There's still one error that makes CI fail, though:

```
./updateHostsFile.py:1094:13: F523 '...'.format(...) has unused arguments at position(s): 0
```


StevenBlack:
Thanks for taking initiative and jumping in, @XhmikosR!

MErging.

PR #1296, opened 2020-05-27
Always use LF in README files

XhmikosR:
On Windows the README files ended up using CRLF. This patch fixes the issue.


StevenBlack:
Thanks again @XhmikosR.  Windows, ugh.  I'm grateful to contributors who watch the Windows-side of things.

Merging.

PR #1302, opened 2020-06-05
Add blasze IP logger links

jnguyen1098:
I noticed the `hosts` file blocks `grabify.link` and `iplogger.org`, two IP loggers, but it doesn't block `blasze.tk` or `blasze.com`, which is also an IP logging service.

### DIFF
--- a/hosts
+++ b/hosts
@@ -2799,7 +2799,13 @@ ff02::3 ip6-allhosts
 0.0.0.0 www.govrefund-uk.com
 
 # Added May 26, 2020
-0.0.0.0 seemascakes.com # End StevenBlack
+0.0.0.0 seemascakes.com
+
+# Added Jun 5, 2020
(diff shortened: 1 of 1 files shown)


StevenBlack:
Hi Jason @jnguyen1098, thank you for this contribution.

Unfortunately you applied this change to a ***generated*** file.  Changes should be made in a source file, either in a [data](https://github.com/StevenBlack/hosts/tree/master/data) source or an [extension source](https://github.com/StevenBlack/hosts/tree/master/extensions) each of which has a different origin.  See the readme for more details on each of these source files.

I suggest you add this domain to [my personal ad-hoc hosts list](https://github.com/StevenBlack/hosts/blob/master/data/StevenBlack/hosts) and, from there, it will get added to our various amalgamated hosts files for the next release.

Closing.

PR #1305, opened 2020-06-06
Add blasze.tk / blasze.com IP logger links

jnguyen1098:
Hello, I added the domains to the right source files.


StevenBlack:
Thanks Jason @jnguyen1098 that's good to merge!

PR #1308, opened 2020-06-07
Update CI config

XhmikosR:
* update to `actions/cache@v2`
* change cache key to take into account requirements.txt

CC @StevenBlack: should speed things up a bit since `actions/cache@v2` should be faster :)


StevenBlack:
Oh, awesome!  Thanks @XhmikosR!

Merging!

PR #1321, opened 2020-06-24
add link.getmailspring.com

taobojlen:
[Mailspring](https://getmailspring.com/) is an email client that lets users add tracking pixels and link redirects to keep tabs on who opens their emails.

Tracking happens through the domain `link.getmailspring.com` which I've added here.


StevenBlack:
Thanks Tao @taobojlen, good catch.

Merging.

PR #1333, opened 2020-07-08
Update readme_template.md

XhmikosR:
I'm not familiar with NixOS, but this seems more appropriate.

/CC @sheepfleece


StevenBlack:
Looks good to me.

Will merge this if there are no objections.

Thanks @XhmikosR!

StevenBlack:
Thanks Nissar @funilrys. 

Merging.

PR #1339, opened 2020-07-10
Check if a service is running (Fixed #1337)

### LINKED ISSUE #1337: Add check for running DNS cache services (updateHostsFile.py)
Currently the script restarts any service it finds available resulting in a buggy behaviour simply because performing:
`sudo systemctl restart dnsmasq.service` will also start stopped (and or disabled) `dnsmasq` service - which is often undesired.

**TL; DR;** Before restarting a systemctl service check if its already running.

codeswhite:


StevenBlack:
Thank you @codeswhite!  Good addition.

Merging.

PR #1342, opened 2020-07-10
Implemented connectivity check (Fixed #1038)

### LINKED ISSUE #1038: Add retry option for updateHostsFile.py
Sometimes I get a network timeout when running `python updateHostsFile.py -as`

Please add an option to retry downloading from the source when that happens instead of just skipping it.

codeswhite:


updateHostsFile.py
```diff
@@ -325,6 +326,34 @@ def main():
...
+    -> bool
+        True : A ping succeded
+        False : All retries have failed
+    """
+    while retries:
+        if 0 == subprocess.call(['ping', 'one.one.one.one', '-c', '1'],
+                                stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL):
+            return True
+
+        print(f'No internet connection! Retrying in {delay} seconds')
```
funilrys, line 352: We support Python 3.5+ so please avoid f-string.

updateHostsFile.py
```diff
@@ -325,6 +326,34 @@ def main():
...
+        How many seconds to wait before retrying again (on ping failure)
+
+    Returns
+    -------
+    -> bool
+        True : A ping succeded
+        False : All retries have failed
+    """
+    while retries:
+        if 0 == subprocess.call(['ping', 'one.one.one.one', '-c', '1'],
```
funilrys, line 348: The script should run on windows too.

Please assume that on windows `ping` will expect `-n` instead of `-c`.

So.
```python
import platform

if platform.system().lower() == "windows":
     count_arg = "-n"
else:
     count_arg = "-c"
```

codeswhite:
I'm happy to help!
I have reverted the previous commit as a result of your clarification:
> You should take into consideration that the timeout/network error mentioned by OP may happen later in the process!

---

Only after pushing I noticed your note: 
> ... this should be implemented into the loop of the `update_all_sources()` function.

I believe handling `name resolution` exception right after it occurs in `get_file_by_url()` is a better practice as it prevents `too broad exception clause` in `update_all_sources()`
See [this answer](https://stackoverflow.com/a/30442821) for more info.

---

> The script should run on windows too.

This solution should be good to go (unfortunately I use no windows so cannot test).
Although I'm afraid that on uncommon OSes (e.g. OpenBSD) the exception message might be different thus breaking the code, anyone with an OpenBSD to test it?

updateHostsFile.py
```diff
@@ -1482,12 +1482,17 @@ def get_file_by_url(url):
...
+        try:
+            with urlopen(url) as f:
+                soup = BeautifulSoup(f.read(), "lxml").get_text()
+                return "\n".join(list(map(domain_to_idna, soup.split("\n"))))
+        except Exception as e:
+            if 'failure in name resolution' in str(e):
+                print('No internet connection! Retrying in {} seconds'.format(delay))
+                time.sleep(delay)
+                retries -= 1
+            return print("Problem getting file: ", url)
```
funilrys, line 1495: @codeswhite, you are not retrying with your `return` statement.

1. This return statement combined with the `print()` function might be obscure for future python contributors. Let's keep it in the same format as before :wink: 
2. This return statement should be outside of the `while` statement if we want to retry.

In the current form and in my understanding, you are not retrying.
Instead, you are handling the resolution issue, waiting for 10 seconds and then return `None`. 

Isn't it better to add a new exception handler? Indeed, we could have 2 handlers. A more general `Exception` for everything we don't necessarily want to control and the `urllib.error.URLError` exception with the `failure in name resolution` message which will handler our delay.

Please let me know if I missed something or if my understanding late at night (Berlin Time) is wrong.

codeswhite:
Generally its better to use multiple exception handlers in similar situations (where multiple exceptions may occur) but in our case it would go against the existing code patterns of the script.

For now I wanted to follow the existing code patterns and use the same print statement for various exceptions including when maximum retries have been reached.
It seem that in our scenario using continue and break inside the `while` loop is the best practice.
The `continue` statement "makes us retry" by returning code flow to the beginning of the while loop,
And the `break` statement ensures all other exceptions won't invoke a retry but instead exit the loop unconditionally.

I want to open an issue about refactoring the code and 'migrating' to better exception handling (possibly including definition of custom `Exception` sub-classes).

funilrys:
Thanks for your input @codeswhite ! Exactly where we should go.

It's yours Steve @StevenBlack!

StevenBlack:
Thanks @codeswhite and thanks Nissar @funilrys for the careful review.

Merging.

PR #1355, opened 2020-07-19
Remove miniconda and add Windows CI support

XhmikosR:
I'm not a Python/pip expert so if there's a safer operator we could use in requirements.txt let me know. I feel like `>=` is too lax.

Notes:

- I used `python` so that it works on every OS; `python3` didn't work on Windows
- This should result in faster CI runs because we install less stuff and most of the time they will be cached

/CC @StevenBlack @funilrys for review :)


requirements.txt
```diff
@@ -1,2 +1,3 @@
 lxml>=4.2.4
 beautifulsoup4>=4.6.1
+flake8>=3.8.3
```
StevenBlack, line 3: @XhmikosR I think we could say...

```
lxml>=4.2.4,<=5.0
beautifulsoup4>=4.6.1,<=5.0
flake8>=3.8,<=4.0
```

AlexanderCecile:
Miniconda wasn't working on Windows?

requirements.txt
funilrys: Correct @StevenBlack.

/cc @XhmikosR

XhmikosR:
Please check again @StevenBlack @funilrys

StevenBlack:
Thanks @XhmikosR and Nissar @funilrys.

Merging.

PR #1356, opened 2020-07-20
sudo provides root privileges

ngaro:
`sudo` without special options provides the user with root privileges so he doesn't need them before running the commands.

### DIFF
--- a/alternates/fakenews-gambling-porn-social/readme.md
+++ b/alternates/fakenews-gambling-porn-social/readme.md
@@ -381,7 +381,7 @@ sc stop "Dnscache"
 
 ### Linux
 
-Open a Terminal and run with root privileges:
+Open a Terminal and run:
 
 **Debian/Ubuntu** `sudo service network-manager restart`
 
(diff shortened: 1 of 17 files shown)


ngaro:
I just noticed that there is more then 1 readme file, this commit updates all of them

StevenBlack:
Hi Nikolas @ngaro thank you for this.

Our `readme.md` files are generated by python scripts included in this repo.  The file [readme_template.md](https://github.com/StevenBlack/hosts/blob/master/readme_template.md) is where this change should be made, just once 😄 

Closing.

ngaro:
I noticed this also, but since all generated `readme.md`-s are also in the repo this is more correct

StevenBlack:
I'm sorry, what do you mean?  It's possible you don't understand because maybe I was unclear?

All the `readme.md` files are products of a generative process.

All the readme files get destroyed and regenerated with each release.  Changes to the readme need to go into `readme_template.md`.

ngaro:
I mean that if these `readme.md` files are generated by a script then they shouldn't be in the git repo. Because they are the in the repo changing only the template will leave non-compatible files behind.

StevenBlack:
@ngaro, thank you for your input.  I'm going to lock this thread now.

PR #1358, opened 2020-07-30
Add new domain servenobid.com

bigdargon:
I see this ad domain has not been added. This domain does not have a server in Vietnam, so I created this PR to add.


StevenBlack:
Thank you @bigdargon this is a good addition.

Merging!

PR #1364, opened 2020-08-04
Adding Microsoft telemetry domains

Keating950:
This PR adds some of the telemetry endpoints that Windows 10 [attempts to prevent you from blocking](https://www.bleepingcomputer.com/news/microsoft/windows-10-hosts-file-blocking-telemetry-is-now-flagged-as-a-risk/) via a HOSTS file.

### DIFF
--- a/data/StevenBlack/hosts
+++ b/data/StevenBlack/hosts
@@ -2915,3 +2915,10 @@
 
 # Added August 3, 2020
 0.0.0.0 instui201.info
+
+# Added August 4, 2020
+0.0.0.0 telemetry.remoteapp.windowsazure.com
+0.0.0.0 telemetry.urs.microsoft.com
+0.0.0.0 us.vortex-win.data.microsoft.com
+0.0.0.0 watson.ppe.telemetry.microsoft.com
+


StevenBlack:
Hello @Keating950 please review [our many past discussions of this](https://github.com/StevenBlack/hosts/issues?q=is%3Aissue+Microsoft+telemetry+is%3Aclosed).

If you still feel we should add these, then there is also the merge conflict described above ⬆️ .

Microsoft Windows is a shitshow.   I'm open to considering this again, but I fear it just brings an avalanche of complaints from mostly non-technical Windows users.

Input from others is welcome on this — I'm a bit battle weary TBH.  Dan @dnmTX?  Nissar @funilrys?  LE @llacb47? @anudeepND? @xetorixik?

StevenBlack:
See also Issue #1365 ***Windows Defender reporting telemetry-blocking hosts file is a risk***

dnmTX:
It's wildly known by now that that Microsoft is bypassing the hosts file to report their tracking/telemetry and god knows only what other garbage to their servers.
Smart Windows users who care about their privacy already moved to many available third-party solutions(Pi-Hole,routers etc.)
If whoever still using the Windows hosts file and thinks he's protected, i'd say: GO SEE A DOCTOR!
Steve @StevenBlack i can only think of one solution:
Clear out all the Microsoft entries from your list(less crybabies for sure) and instead use [@crazy-max's list](https://raw.githubusercontent.com/crazy-max/WindowsSpyBlocker/master/data/hosts/spy.txt) as a extension with a BIG DISCLAMER that it's **OPTIONAL**.
With @crazy-max's blessing of course 😉

bremich:
There might also be issues with various MS services seemingly unrelated to telemetry. @Bryantdl7 has noted elsewhere that [there are multiple telemetry domains which stop Xbox achievements from working](https://github.com/lightswitch05/hosts/issues/161). 

While none of these hosts are present in this pull request, blocking MS telemetry domains might have other unintended consequences. 

Having a separate variant of the blocklist as @dnmTX suggested is probably the best solution to both have an option of blocking then and avoid an influx of issues.

Keating950:
@bremich's idea seems very reasonable to me.

PR #520, opened 2018-02-28
Possible fix of the encoding and/or downlaod issue(s)

funilrys:
Hello, to all member of this community!

After a long time of commenting, watching, reading and learning from issues posted here it's time for me to propose the following patches/commits.

It's been a long time since I wanted to search a possible solution to this recurrent issue and here's my proposition.

Basically, since I consider that many encoding issues are because upstream maintainer does not convert/encode the domains which rely on [IDN homograph attack](https://en.wikipedia.org/wiki/IDN_homograph_attack), I decided to read all line and convert/encode all domain using `str.encode('IDNA')` (I do not change original structure). 

Because my main objective was to fix @notDavid's [protocol](https://github.com/StevenBlack/hosts/issues/514#issuecomment-368932152) which is the only one I was able to reproduce, I introduced `BeautifulSoup()` to do the decoding of the downloaded data before we read each line and encode/decode.

## About the tests

Before anything, sorry for adding the `fix tests issues` commits but I wanted to fix those before submitting my PR. You can find the tests I used as the reference to committing those change at https://travis-ci.org/funilrys/hosts.

Please note that I have to submit an issue to upstream but if you see the following or related when runing the tests under `Python >= 3` it's not a local issue. 
```
/home/travis/miniconda3/envs/hosts/lib/python3.5/site-packages/bs4/builder/_lxml.py:250: DeprecationWarning: inspect.getargspec() is deprecated, use inspect.signature() instead
```

## About `requirements*.txt`

This submission also introduces `requirements.txt` files which I generally use to install dependencies with `pip install -r requirements.txt` (I'm almost everytime under `virtualenv`).

As I did not get an answer ([cf](https://github.com/StevenBlack/hosts/issues/514#issuecomment-368983762)) I took the initiative to includes them anyway.

Please note that the `requirements*.txt` files were generated by [pipreqs](https://github.com/bndr/pipreqs).

## So to quote my commit message

Please note that this patch also introduces `domain_to_idna()`
which is in charge of converting a domain in a line into
IDNA and/or UTF-8 format.

Also, note the introduction of BeautifulSoup() which helps
us to decode data from the downloaded URL.

Fixes (issue(s)/protocol(s) I was able to reproduce):
  * https://github.com/StevenBlack/hosts/issues/514#issuecomment-368932152
    
Possible fix of (issue(s)/protocol(s) I wasn't able to reproduce):
  * https://github.com/StevenBlack/hosts/issues/514#issue-300048106
[...]

PR #1392, opened 2020-08-21
Introduce the Requests library

AlexanderCecile:
Before anything else, I would like to apologize for the somewhat awkward/bad first issue I had opened. I think this is much closer to being up to standards, I have done a good amount of digging through previous issues and changes.

&nbsp;

**Introduction**

BeautifulSoup4 (bs4) was added as part of PR #520, in order to automatically detect the correct encoding to use when fetching a file using the urllib.requests module. 

To quote the [urllib.requests](https://docs.python.org/3/library/urllib.request.html#module-urllib.request) documentation:

> The [Requests package](https://requests.readthedocs.io/en/master/) is recommended for a higher-level HTTP client interface.

I believe that using Requests, which includes mechanisms for detecting encoding, would provide a few solid benefits over the combination of bs4 and urllib.requests.

&nbsp;

**A few benefits**

First, it is worth mentioning that Requests uses chardet to detect encoding, which is an optional dependency for bs4 in order to improve the quality of the guess (see [here](https://www.crummy.com/software/BeautifulSoup/bs4/doc/#unicode-dammit)). As such, I would expect Requests to perform at least as well, if not better, than bs4 in this regard.

The fact that bs4 is an HTML parser can lead to some less-than-ideal behaviour. As an example, take one of the hosts files used in this project, at https://someonewhocares.org/hosts/zero/.

Here are the first few hostnames:
```
#<localhost>
127.0.0.1	localhost
127.0.0.1	localhost.localdomain
255.255.255.255	broadcasthost
::1		localhost
127.0.0.1	local
::1		ip6-localhost ip6-loopback
fe00::0		ip6-localnet
ff00::0		ip6-mcastprefix
ff02::1		ip6-allnodes
ff02::2		ip6-allrouters
ff02::3		ip6-allhosts
#fe80::1%lo0	localhost
#</localhost>
```

Here is how those same lines appear in the combined hosts file:

https://github.com/StevenBlack/hosts/blob/2814997df51466f051ddf2f9c50fe437a62d5c97/data/someonewhocares.org/hosts#L69-L82

Hint: Look at the first and last lines.

Some debugging confirms that bs4 treats `<localhost>` and `</localhost>` as HTML tags, and they disappear when we retrieve the content of the page using the `get_text()` method [here](https://github.com/StevenBlack/hosts/blob/2814997df51466f051ddf2f9c50fe437a62d5c97/updateHostsFile.py#L1496).

Switching to Requests allows us to drop two dependencies, which is especially good when you consider that the vast majority of the functionality they provide was going unused anyway. The higher-level interface it offers should also help simplify the code.

&nbsp;

I don't have much experience with this kind of thing, so please let me know if I need to explain anything further, or if I have missed something.

Thank you for your consideration.


updateHostsFile.py
```diff
@@ -21,15 +21,12 @@
 import time
 from glob import glob
 
-import lxml  # noqa: F401
-from bs4 import BeautifulSoup
+import requests
```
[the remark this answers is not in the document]
AlexanderCecile: > My mind just blocked on the fact that `requests` is listed in `requirements.txt` 😂

Wait, did I mess something up there?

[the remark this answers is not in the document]
gfyoung: No, you're fine here.  This thread can be considered resolved. 👍

updateHostsFile.py
```diff
@@ -1469,40 +1466,27 @@ def maybe_copy_example_file(file_path):
...
+
+    Parameters are passed to the requests.get() function.
 
     Parameters
     ----------
-    url : str
-        The URL at which to access the data.
+    url
+    params
+    kwargs
```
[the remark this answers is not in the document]
AlexanderCecile: @gfyoung I updated the documentation, and re-added the returns section, let me know what you think.

updateHostsFile.py
```diff
@@ -1469,40 +1466,27 @@ def maybe_copy_example_file(file_path):
...
-                print('No internet connection! Retrying in {} seconds'.format(delay))
-                time.sleep(delay)
-                retries -= 1
-                continue
-            break
-    print("Problem getting file: ", url)
+    req = requests.get(url=url, params=params, **kwargs)
+    req.encoding = req.apparent_encoding
+    res_text = "\n".join([domain_to_idna(line) for line in req.text.split("\n")])
+    return res_text
```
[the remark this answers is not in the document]
AlexanderCecile: I made some changes, it should now catch any requests-related exceptions, and decoding errors.

updateHostsFile.py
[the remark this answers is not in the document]
gfyoung: These look good!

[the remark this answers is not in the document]
AlexanderCecile: I'll see if anything can be improved, particularly when it comes to the parameter types. Requests uses a different convention and my IDE seems to be getting a bit confused when inferring from the new docstring.

testUpdateHostsFile.py
```diff
@@ -1615,47 +1614,6 @@ def test_line_without_prefix(self):
...
-    def test_read_url_decode_fail(self, _):
-        url = b"www.google.com"
-        self.assertIsNone(get_file_by_url(url, delay=0))
-
-        expected = "Problem getting file:"
-        output = sys.stdout.getvalue()
-
-        self.assertIn(expected, output)
-
-
```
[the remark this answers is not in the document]
AlexanderCecile: I added two basic tests.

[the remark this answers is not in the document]
AlexanderCecile: I added two basic tests to start off with.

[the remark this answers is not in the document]
gfyoung: Looks like a good start to me 👍

StevenBlack:
Thank you @AlexanderCecile, since this nominally breaks current installations with the following...

```bash
$ python3 testUpdateHostsFile.py
Traceback (most recent call last):
  File "testUpdateHostsFile.py", line 20, in <module>
    import requests
ModuleNotFoundError: No module named 'requests'
```

... this needs to come in a major release.

Unless we auto-fix this, with a check for `requests` and auto install it if it's missing?  Given the mess that is Python packages, I sense this might be doable, but morally reprehensible, perhaps?  This would apply to `testUpdateHostsFile.py` as well as `UpdateHostsFile.py`.

I presently have a preference for auto-healing it, because a major release also poses a question, what else should/could be included? which may mean delay rolling this out.

Also thinking, I wonder if we should spin `requirements.txt` and auto-check for all dependencies, not just `requests`?

**In summary** the choice is, 1) new branch here named `v.3.0`, or patch-in an auto-check and auto-upgrader.

@gfyoung and Nissar @funilrys any thoughts?   My best guess is, new branch for `v3.0`.  if so, then I'll create an issue to track its additional features.

StevenBlack:
@AlexanderCecile @gfyoung @funilrys I can also imagine, just make the next version `v3.0` and declare victory, end of story.

gfyoung:
@StevenBlack : Did you run `pip install -r requirements.txt` ?

@AlexanderCecile updated that file with the `requests` dependency.

AlexanderCecile:
> Unless we auto-fix this, with a check for requests and auto install it if it's missing? Given the mess that is Python packages, I sense this might be doable, but morally reprehensible, perhaps?

Automatically installing the package(s) would be unusual, I don't think it's worth it, especially considering that installing them using the `requirements.txt` is straightforward.


> I presently have a preference for auto-healing it, because a major release also poses a question, what else should/could be included? which may mean delay rolling this out.

I was actually planning a few more things after this feature, so that's might be a good idea. It depends on what else is going with the project, though.

StevenBlack:
@gfyoung oh yes I did!  That's not the issue.  

The issue is, this is a backwards-incompatible change, which has implications for [semver](https://semver.org/).  Also, there is potential for an avalanche of support issues from folks doing `git pull` and, all of a sudden, they can't build.

So the decision tree appears to be

1. Declare `v3.0` and either
  a. declare victory, or
  b. add features because `v3.0` could include other things too, OR

2. make it so `requests` appears magically resulting from our own guard code.

gfyoung:
@StevenBlack : Ah, okay, understood.

Here's a proposal:

1) Do a `try-except` around `import requests` and raise `ImportError` with a message saying that it's required.

2) Declare victory with `v3.0`

WDYT?

AlexanderCecile:
@gfyoung Is 1) really necessary? If the library isn't installed you'll get a `ModuleNotFoundError` anyway, which I believe is also the more appropriate exception for this.

gfyoung:
@AlexanderCecile :

* Meant to write `ModuleNotFoundError` instead of `ImportError` - sorry for the confusion.

* Yes, re-raising the error with a user-friendly error message is preferable.  The reason is that it shows to the end user that this error was anticipated (since this is going to be a breaking change), not unexpected, and will reduce the number of issues raised against this repository about it.

funilrys:
I agree with @gfyoung. A more user-friendly error is preferable if we don't want to deal with that issue every few hours/days.

StevenBlack:
Thanks Alexander @AlexanderCecile I know this might seem like more than you bargained-for.  This is a good PR, but it's a needed foundational change.  That always brings more because of edge cases, and the installed base.  Once we wrap/trap the error, I'll merge this.

AlexanderCecile:
I hadn't considered the user-friendliness angle, that's a good point. It might be a good idea to present a brief message, with a link to a page with more details. Will the change be announced/posted anywhere?


> Thanks Alexander @AlexanderCecile I know this might seem like more than you bargained-for. This is a good PR, but it's a needed foundational change. That always brings more because of edge cases, and the installed base. Once we wrap/trap the error, I'll merge this.

No worries, this is about what I was aiming for anyway: Not a huge change, but not a trivial tweak either.

updateHostsFile.py
```diff
@@ -21,15 +21,16 @@
 import time
 from glob import glob
 
-import lxml  # noqa: F401
-from bs4 import BeautifulSoup
+try:
+    import requests
+except ModuleNotFoundError as e:
```
gfyoung, line 26:
```suggestion
except ModuleNotFoundError:
```

I think this should fix the linting errors.

updateHostsFile.py
```diff
@@ -21,15 +21,16 @@
...
+    raise ModuleNotFoundError("This project's dependencies have changed. The Requests library ("
+                              "https://requests.readthedocs.io/en/master/) is now required.")
```
gfyoung, line 28: This feels a little verbose, but that's fine. 👍

updateHostsFile.py
AlexanderCecile: Oops I hadn't refreshed the page, it should be fixed now.

updateHostsFile.py
AlexanderCecile: How about "The Requests library (https://requests.readthedocs.io/en/master/) is now a required dependency." ? Is "a required dependency" a bit awkward?

AlexanderCecile:
Any idea why the linter is unhappy (https://github.com/StevenBlack/hosts/actions/runs/227972635) ? The issue is with Python 3.5 only, it works locally.

AlexanderCecile:
`ModuleNotFoundError` was apparently added in Python 3.6 ([docs](https://docs.python.org/3.8/library/exceptions.html#ModuleNotFoundError)), which is why [flake8 complains](https://github.com/StevenBlack/hosts/pull/1392/checks), yet it works fine locally with Python 3.5. Hmm.

AlexanderCecile:
Alright, I'm sure I've made a real mess of the git history. 

At least i've found a few things which might make it easier in the future (https://fle.github.io/git-tip-keep-your-branch-clean-with-fixup-and-autosquash.html) ...

updateHostsFile.py
```diff
@@ -21,17 +21,20 @@
...
-else:
+if not PY3:
     raise Exception("We do not support Python 2 anymore.")
 
+
+try:
+    import requests
+except ModuleNotFoundError:  # noqa: F821
+    raise ModuleNotFoundError("This project's dependencies have changed. The Requests library ("  # noqa: F821
+                              "https://requests.readthedocs.io/en/master/) is now required.")
```
gfyoung, line 35: If this error class was introduced in 3.6, then we should `except` on `ImportError` (`ModuleNotFoundError` is a sub-class) and raise that instead of `ModuleNotFoundError`.  Otherwise, 3.5 will break with a `NameError`.

AlexanderCecile: Indeed, it was behaving strangely locally so I thought it wouldn't be an issue. Would it be better to raise different errors based on whether the version is > 3.5, or just use an `ImportError` regardless?

gfyoung: Keep `ImportError` regardless.  This doesn't need to be super elaborate (KISS).

AlexanderCecile: I don't think it would work anyway, so nvm. It all works out in the end!

gfyoung:
The `try-except` looks good.  I think if we could expand the tests to include coverage for failure handling, that would be good.

Then, last but not least, I think this whole chunk of code (https://github.com/StevenBlack/hosts/blob/59ddd34d0edce556289bb9bdf8f1cabe1e41014b/testUpdateHostsFile.py#L1412-L1481) can be deleted.

AlexanderCecile:
> I think if we could expand the tests to include coverage for failure handling, that would be good.

You're referring to the tests for `get_file_by_url()`, right?

gfyoung:
> You're referring to the tests for get_file_by_url(), right?

Yes, that's right.

AlexanderCecile:
Are there any exceptions in particular I should be targeting?

gfyoung:
> On another note, has anyone here used pytest? It seems quite convenient, compared to unittest.

`pytest` is quite useful, though the porting effort would be pretty big.

testUpdateHostsFile.py
```diff
@@ -1616,44 +1547,43 @@ def test_line_without_prefix(self):
...
+        printed_output = sys.stdout.getvalue()
+        self.assertEqual(printed_output, "Error retrieving data from {}\n".format(test_url))
 
-        self.assertIn(expected, output)
+    def test_invalid_url(self):
+        test_url = "http://fe80::5054:ff:fe5a:fc0"
+        return_value = get_file_by_url(test_url)
+        self.assertIsNone(return_value)
+        printed_output = sys.stdout.getvalue()
+        self.assertEqual(printed_output, "Error retrieving data from {}\n".format(test_url))
```
gfyoung, line 1586: Can we mock the responses instead of using live calls to `request.get` ?

AlexanderCecile: That's probably a good idea, yes. I'll dig through the code for Requests to see if there's a convenient place we can mock.

StevenBlack:
@AlexanderCecile @gfyoung when I merge this, I'll crank the version to `3.0` and we'll just declare victory, no other features squeezed-in.

testUpdateHostsFile.py
AlexanderCecile: I'm struggling to find the best place to patch/mock, do you have any ideas? If I go too low, I increase the chances that the tests will break on a different OS, Python version, etc. It seems pointless to go much higher, though, since the code is trivial anyway, and if something fails there then we have much bigger issues to worry about.

gfyoung: Is it possible to mock the `Response` object that you get back from `requests.get` ?

AlexanderCecile: Yes, it's what I'm doing in one of the test cases. The issue is with the ones which are supposed to lead to an exception.

gfyoung: You can mock the `requests.get` call to raise an exception.

AlexanderCecile: Is it even worth it at that point? All we're testing is whether or not the try-except catches the exception.

gfyoung: IMO yes.  `try-except` is still functionality that should be tested.

AlexanderCecile: Done! :)

gfyoung (approved):
@StevenBlack : Back to you

StevenBlack:
Merci Alexander @AlexanderCecile and thank you for the assist, @gfyoung!

Merging.

XhmikosR:
Shouldn't requests version be tightened in requirements.txt? Right now if they release a major version it will be installed, which is not right?

StevenBlack:
@XhmikosR that's a good point!

AlexanderCecile:
@XhmikosR I wasn't sure which version(s) I should limit it to. 

My reasoning was that since we're using only the basic functionality of the library, any version compatible with the Python versions we support should work just fine. We have the tests, and we could keep an eye on the changes in new versions of Requests.

XhmikosR:
It's always better to be safe.

PR #1398, opened 2020-09-01
Update hosts

piluex:
Infomercial site that spreads anti-social conduct and anti-vaccine spam against covid-19 fight in Latin America controlled by USA

### DIFF
--- a/data/StevenBlack/hosts
+++ b/data/StevenBlack/hosts
@@ -2950,4 +2950,5 @@
 0.0.0.0 collect.installeranalytics.com
 
 # Added September 1, 2020
+0.0.0.0 www.infobae.com
 0.0.0.0 infobae.com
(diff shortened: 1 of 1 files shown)


dnmTX:
@piluex almost 3M people on twitter 👇 
Twitter
and 4M+ on Facebook 👇 
Facebook
might disagree with your assessment.
The domain too controversial to be added here,in my opinion.

marktron:
Based on a review of the site's current stories (via Google Translate), nothing seems egregiously malicious to me. Their [wikipedia entry](https://en.wikipedia.org/wiki/Infobae) looks like a straightforward media outlet. I don't think I've seen anything that justifies inclusion in a blocklist.

StevenBlack:
Thanks Mark @marktron.

Closing.

PR #1413, opened 2020-09-30
updateHostsWindows.bat: fix python exe cmd

FadeMind:
Python for Windows using py.exe launcher command and should be used here as well. @StevenBlack please merge.


StevenBlack:
Yup!  Thank you Tomasz @FadeMind.

Merging.

PR #1428, opened 2020-10-12
Removed several entries for bias and censorship

LuckySandwich:
Good filter list overall, but some of the entries are not justified, and were obviously added to the list due to political bias and/or deliberate censorship.  I only briefly glanced through the list, so I'm sure there are more that I missed.

Removed a few that I strongly believe **do not** belong on a block list.  Opposing views, no matter how much you may dislike them, deserve to be heard.  Ideological censorship is just as bad, if not worse, than fake news.


StevenBlack:
Hello @LuckySandwich thank you for your suggestion.

It's evident you don't understand how this repository works, so let me fill you in.

Please submit this request to [Our fake news curator](https://github.com/marktron/fakenews/issues).  Good luck with that.  

Closing.

PR #1445, opened 2020-10-31
CI: add Python 3.9

XhmikosR:
@StevenBlack does Python have LTS releases? If so, maybe it would make more sense to reduce the CI-tested versions to the LTS ones?


StevenBlack:
@XhmikosR σας ευχαριστώ!

Issue #235, opened 2017-01-17
Fakenews


iandoug:
hi

I think the FakeNews lists should be in a separate file ... not everyone is going to agree on what is fake news and what is not. While the intention is fine, at the moment the long embedded lists are verging on Net Nanny or censorship .... IMHO :-)

thanks.

Jezza-x3:
The fakenews list is maintained by a user other than Steven Black. 

Steven has made it clear that the person has good intentions and he will keep it here. 

You DO NOT have the get one with fake news blocked. You have a choose the hosts you will get. If you choose it, it is your fault. 

See issue #232

iandoug:
Hi
Excuse my ignorance. I came here from 
https://www.darrentoback.com/this-script-updates-hosts-files-using-a-multi-source-unified-block-list-with-whitelisting

which I understood as this project being a python script to combine a bunch of blacklists into one massive hosts file. Which I took as meaning "needs all the lists in the repo". After your comment I took a closer look and now see that I should not take all the lists...

I think it would be clearer if the readme spelled that out, or if for example each catogory (malware, ads, porn, etc) had its own list, and then a separate section of the readme, after a heading like "premerged lists" or somesuch, gave the various pre-merged combos which now exist.

Thanks, Ian

inoas:
@iandoug that confirms my [assumption](https://github.com/StevenBlack/hosts/issues/232#issuecomment-272877491) that people will simply download the hole list.

I do agree that there should be a sane recommended default.

StevenBlack:
Hi @iandoug thanks for stopping-by.

> I think it would be clearer if the readme spelled that out, or if for example each catogory (malware, ads, porn, etc) had its own list, and then a separate section of the readme, after a heading like "premerged lists" or somesuch, gave the various pre-merged combos which now exist.

Up-front, at the [very top of the repo, there is this table](https://github.com/StevenBlack/hosts#unified-hosts-file-with-base-extensions) with links to individual readme for each variant.

I'm open to suggestions to make this more clear.

The default list is Adware + Malware and this is the one most people download as it's the one featured in the root of the repo.

iandoug:
TBH, further to my previous message about your script being a combining / de-duping script, I initially interpreted the list of variants as being a collection from different places, which your script would then reduce down, dedup etc. 

ie I took the variants as "pre-cleanup" rather than "post-cleanup".

kronflux:
[...]

Issue #232, opened 2017-01-16
Consider removing "fake news" list? 


persona5:
I don't like where this is going. 
This used to be the best anti-malware, anti-tracking and adblock hosts list out there, but I'm not sure anymore.
"Fake news" list looks really political and usually, when things get political, shit happens. Smells censor-ish.

People tend to imbue their political views in their works, and I'm concerned that this might happen to this repo. Like, inserting websites you don't agree with into unified hosts base to block them, or something like this. I would prefer to have politics as far away from it as possible. 

Can you reconsider and push out politics away into separate designated repository?

blaqkr:
it's not like you must download the list, that's why it has 16 variants.

StevenBlack:
Yes echoing @blaqkr here... 

Fake news is an *extension* to the base hosts file.  You need to actively download the versions that have this extension mixed-in, or explicitly flag it for inclusion if you are building the hosts file yourself using the Python script.

So it's just like the *porn* or *social media* extensions.  You would include those for specific reasons.  *Fake news*, same thing.

Better?

inoas:
The problem with Fake News is what to categorize. And it isn't so easy. Major news sites have been copying faux/hoax/fake news. Facebook and Twitter are the primary source of distribution for Fake News. So you want to put nyt, wp, fb, tw etc on that list?

StevenBlack:
The curator of Fake News is @marktron who's done, and continues to do, a fine job.  He's a reasonable guy.  He's not curating this because it's easy.

Originally both CNN and Fox News were on the list.  As a Canadian, I feel both should probably be listed, but it's a close-call.   I trust Mark.

inoas:
I don't think its a good idea at all, at least not until all major news sites and social networks are listed because they are the primary fake news distributors.

StevenBlack:
@inoas I'm closing this issue now.  Happy computing!

inoas:
@StevenBlack I do value your work highly, but this crosses the line into dark waters.

Obviously you are the maintainer and all I can and want to do is appeal to your critical thought.
People will download the full package and having things like rt.com, CNN or Fox News on the list really shows the inherit issue with this approach.

You can consider later - I hope!

Carpe Diem :)

marktron:
@inoas & @persona5: I understand your concerns, this is something I've been thinking about a lot.

[...]

Issue #601, opened 2018-05-05
Bwahaha blocks all conservative and russian news as fakenews


Biorazero:
0.0.0.0 ruptly.tv
0.0.0.0 www.infowars.com
0.0.0.0 www.naturalnews.com
0.0.0.0 www.prisonplanet.com
0.0.0.0 www.rt.com
0.0.0.0 www.wnd.com
0.0.0.0 www.zerohedge.com

what a loser

Biorazero:
Don't you think you should let your users know that the context of fakenews you're using is coming from a heavily biased viewpoint. If you only added clickbait sites such as yournewswire.com and others to the list it would have been different, but you added mainline ones like wnd and zerohedge, you're out of your effing mind.

What about this hosts file:

0.0.0.0 stevenblack.com

Biorazero:
Is your so called "reputable source" snopes.com for example, a couple with a cat in a dorm room.

dnmTX:
**NOTHING TO SEE FOLKS......JUST ANOTHER "CEO" BLOWING UP STEAM**

Biorazero:
Allright, I may have been too quick to blame the repo owner, it's a consolidation from other curators, I apologize. But still, he seems to agree with it and going along with it so far.

raziel-:
@Biorazero
You can whitelist anything, or you can completely skip Fakenews

funilrys:
May I ask what is the real object behind this issue? 

If you want to blame, as a question or report an issue regarding the fake news extension please report to https://github.com/marktron/fakenews/issues.

@StevenBlack please close as this issue should be handled by @marktron.

Biorazero:
@funilrys Yeah, I just couldn't get to it in the moment, fakenews has been going on since forever, and it just happens to be a big realization right now, however some are being falsely labeled while others that were fake all these years aren't given enough attention.

For example, there are a lot of youtube channels which literally make money off fake CGI-generated UFOs that don't exist, including taking official video from NASA ISS Livestream and modifying it making it look like that NASA "cut the feed" and inserting objects into the video that weren't seen in the actual livestream, such behavior discredits all the cases where there is a real genuine UFO or something.

[...]

PR #1449, opened 2020-11-04
Clarify unavoidable bias in the fakenews category

ohadschn:
Related to https://github.com/StevenBlack/hosts/issues/232 and per your comments in https://github.com/StevenBlack/hosts/issues/235:
> I'm open to suggestions to make this more clear.

> I'm always happy to improve things. Pull Requests resonate best with me

I've used the language from the `fakenews` repo almost verbatim: https://github.com/marktron/fakenews#personal-bias-disclosure

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -35,6 +35,8 @@ The **Non GitHub mirror** is the link to use for some hosts file managers like
 [Hostsman for Windows](http://www.abelhadigital.com/hostsman/) that don't work
 with GitHub download links.
 
+**Note on fake news** - Defining what is or is not "fake news" in today’s political environment is challenging. The `fakenews` list below is curated by a [single (left-leaning) person](https://github.com/marktron/fakenews#personal-bias-disclosure), and as such may contain sites which would be judged differently by another curator.
+
 Host file recipe | Readme | Raw hosts | Unique domains | Non GitHub mirror
 ---------------- |:------:|:---------:|:--------------:|:-------------:
 Unified hosts = **(adware + malware)** | [Readme](https://github.com/StevenBlack/hosts/blob/master/readme.md) | [link](https://raw.githubusercontent.com/StevenBlack/hosts/master/hosts) | 60,248 | [link](http://sbc.io/hosts/hosts)


StevenBlack:
Denied.

Closing.

ohadschn:
May I ask why? I practically copy-pasted the disclaimer from `fakenews` repo itself...

StevenBlack:
Because I'm not entertaining pull requests that explicitly ascribe motive, or leanings, of list our curators.  That's never going to happen.  Am I clear Ohad @ohadschn?

ohadschn:
But I'm quoting the guy himself, in his own words, clearly pointing out that said leanings affect the list (presumably with no malicious motive)... What if we removed "left-leaning"?

StevenBlack:
Ohad @ohadschn I am not entertaining pull requests to characterize the Fake News extension in any way. I have no caveats to offer, nor am I seeking guidance from anyone on this.  Furthermore our users are certainly not dummies; they don't need any sort of sophomoric caveats.

I understand the cookie doesn't crumble the way you see it, but that's your problem, not mine.  I am completely comfortable, and I have complete confidence in @marktron's integrity to manage and curate this.  

Nobody here is forcing anybody to use the Fake News packages we make possible.  It's optional.  Downstream users may bundle it in packages they distribute.  Your issue is with them, not us.

ohadschn:
I agree that your users, and people in general, are not dummies. As such, personally I don't think the Fake News extension is needed at all, at least not one as ambitious as the current one (rather than a super obvious one banning straight out scam and UFO sites). In fact this has been suggested before: https://github.com/StevenBlack/hosts/issues/232.

Assuming you do want to keep it though, I think a disclaimer does make sense, especially when it's basically a copy-paste of the one the repo's owner put in his repo expressly for this purpose (presumably you do not consider his own wording sophomoric). I do not consider myself a "dummy" to put it mildly, and yet when I originally downloaded the list I did not dream there would come a day it would actually block what I view as a legit (if biased) news site - on purpose!

Yes, the list is optional, you are not forcing anyone, downstream users can change it and add exception, etc. But the fact is, the vast majority of users will just take a quick look, say "hey I can block fake news, cool" and enable it. Then 3 years later when their friend sends them a link they can't open, it's doubtful they would make this connection.

I have a few open source repos and projects of my own. When people open issues I don't just tell them "hey, it works for me, so it's your problem, nobody's putting a gun to your head to use it" - and I'm sure you're not doing the same in the case of technical issues. In my opinion, while this proposal is not technical, it could potentially benefit a large swath of your users (e.g. US has about 50% conservatives). I'm just one example, there are many more in the issues (random example https://github.com/StevenBlack/hosts/issues/601), and I'm sure you want your work to benefit all people, irrespective of their political viewpoint.

What if we added a more generic caveat that didn't only single out the Fake News extension? Something like:
> These lists are curated by independent people, and might contain categorizations which are erroneous or do not match your definition of the categories below. For example, defining what is or is not "fake news" might be subjective, and as such may contain sites which would be judged differently by another curator. The same can be true for gambling, or even pornography.

PR #1453, opened 2020-11-08
Added tinypass cdn domain to hosts file

jo3rg:
block out adblock blocker ping requests from tinypass

### DIFF
--- a/hosts
+++ b/hosts
@@ -14833,8 +14833,6 @@ ff02::3 ip6-allhosts
 
 # [tinypass.com]
 0.0.0.0 cdn.tinypass.com 
-0.0.0.0 api-v3.tinypass.com
-0.0.0.0 api.tinypass.com
 
 # [tiqcdn.com]
 0.0.0.0 tiqcdn.com
--- a/hosts
+++ b/hosts
@@ -14831,6 +14831,9 @@ ff02::3 ip6-allhosts
 # [tingyun.com]
 0.0.0.0 beacon.tingyun.com
 
+# [tinypass.com]
+0.0.0.0 cdn.tinypass.com 
+
 # [tiqcdn.com]
 0.0.0.0 tiqcdn.com
 0.0.0.0 akamai.tiqcdn.com
--- a/hosts
+++ b/hosts
@@ -14833,6 +14833,8 @@ ff02::3 ip6-allhosts
 
 # [tinypass.com]
 0.0.0.0 cdn.tinypass.com 
+0.0.0.0 api-v3.tinypass.com
+0.0.0.0 api.tinypass.com
 
 # [tiqcdn.com]
 0.0.0.0 tiqcdn.com


StevenBlack:
Hi @jo3rg can you please clarify?  I'm uncertain how to interpret the description above ⬆️⬆️

jo3rg:
@StevenBlack Tinypass is one of the big paywall firms and has implemented it's adblock blocking mechanism using the following JS snippet: https://cdn.tinypass.com/api/tinypass.min.js

dnmTX:
`0.0.0.0 api-v3.tinypass.com`
`0.0.0.0 api.tinypass.com`

jo3rg:
done and sorry for the messup with git users :)

llacb47:
Bad idea, it will break subscriptions and login on news sites.

https://github.com/AdAway/adaway.github.io/issues/75
https://github.com/AdAway/adaway.github.io/issues/96
https://github.com/ryanbr/fanboy-adblock/issues/1417
https://github.com/ryanbr/fanboy-adblock/issues/1423

jo3rg:
@llacb47 my PR is not that big, it consists of cdn and api hosts.
I think it would have been enough if we would remove `id` and `buy` subdomains from our list:
https://github.com/AdAway/adaway.github.io/issues/96

StevenBlack:
Thank you, as always, for the assists Dan @dnmTX and LE @llacb47.

I'm always wary of blocking domains that identify as `cdn` — content delivery networks.  We wouldn't block Amazon AWS for example, only very specific subdomains.

@jo3rg one more detail.  Sorry for not noticing before — the Github mobile app is not always clear about file paths. The `/hosts` file here is generative; it's a product of the host file amalgamation process.  **Therefore** these changes need to be made to `/data/StevenBlack/hosts` which is a component used in the build process.

Bit I think blocking `cdn.tinypass.com` will break too many things.

jo3rg:
@StevenBlack ok I see will do the change as soon as somebody has verified if it will break anything.
But from my point of view it won't break anything, I've event tried the checkout process mentioned in:
https://github.com/AdAway/adaway.github.io/issues/96
@dnmTX & @llacb47 can you please verify?

dnmTX:
I didn't know that there were so many discussions @AdAway about it and i trust @jawz101 judgement when it comes to removal of domains from his lists. I'm strictly against blocking domains that break any website functionality(that's where AdBlocker comes handy). So i agree with Steven @StevenBlack  and LE @llacb47 for not adding those domains here. No need to move same old issues from @AdAway to here.

StevenBlack:
Yup, that's a good take Dan @dnmTX.

Thank you for the suggestion @jo3rg but we'll pass on this because of possible side-effects.

Closing.

jo3rg:
blocking `cdn.tinypass.com` only as initially requested and tested by me shouldn't have any sideeffects,
but it's okay, i'll add it to my local exceptions :/

PR #1477, opened 2020-12-04
add immopreneur.de and subdomains

breucode:
immopreneur.de (probably hacked) is regularly at the top of many search results in google and redirects to scam sites like slowdistanttube3.live

### DIFF
--- a/data/StevenBlack/hosts
+++ b/data/StevenBlack/hosts
@@ -2844,3 +2844,11 @@
 # Added November 22, 2020
 0.0.0.0 pamsg.online
 0.0.0.0 www.pamsg.online
+
+# Added December 4, 2020
+0.0.0.0 immopreneur.de
+0.0.0.0 lp.immopreneur.de
+0.0.0.0 campus.immopreneur.de
+0.0.0.0 affiliates.immopreneur.de
+0.0.0.0 api.immopreneur.de
+0.0.0.0 www.immopreneur.de


dnmTX:
VirusTotal agree with you.I also checked related subdomains and their IP's are riddled with malware domains so it's probably a good idea all to be added here:
`0.0.0.0 lp.immopreneur.de`
`0.0.0.0 campus.immopreneur.de`
`0.0.0.0 affiliates.immopreneur.de`
`0.0.0.0 api.immopreneur.de`
0.0.0.0 www.immopreneur.de`

I'm not sure about the `*.live` domain(it's malware for sure). They all are short lived and replaced with new once instantly.
Here is a screenshot from the `*.live` domain IP:

VirusTotal

..and a comment from some security guru:

VirusTotal1

StevenBlack:
Thanks @breucode — these changes are in the wrong file.  The `/hosts` file is generated from data.  This change needs to be applied to `/data/StevenBlack/hosts`.

Please include domains listed by @dnmTX.  Thanks!

breucode:
I just added immopreneur.de and their subdomains to the correct file. Thanks @StevenBlack and @dnmTX

StevenBlack:
Dan @dnmTX is the company home page toxic, in your view?    www.immopreneur.de

I see 77 `http` requests by that home page, but nothing toxic-looking *per se*, there.

This is in keeping with blocking the delivery vectors, but not the company home page, if the homepage isn't a vector.  The same question would apply to api.immopreneur.de, too.

dnmTX:
Shoot 😞. The site is **legit**(didn't bother checking it earlier as i'm temporarily without AntiVirus,sorry). 
That is probably a third-party cookie does that redirect.
@breucode try to reproduce by blocking the third-party cookies in your browser(if unsuccessful block all the cookies to that website) and post the results.
Steve @StevenBlack sorry again,let's get to the bottom of this before you take any decisions 👍

breucode:
The website I accessed via google was https://immopreneur.de/forum/topic/4558-stellplatz-nicht-im-grundbuch-eingetragen/ .

Now, when I access it via google, or directly, without deleting cookies since I opened this issue, the site works properly, so I have no clue, why that happened...

dnmTX:
I see no ads or redirects either so i think it's safe to say that it was a false alarm and the issue can be close now.

StevenBlack:
Thank you everybody, @breucode and Dan @dnmTX.

Closing.

PR #1508, opened 2021-01-19
Adware domain: resumersvo.fun

FadeMind:
Adware domain. Blocked in default by ESET.


StevenBlack:
Thank you Tomasz @FadeMind.  Merging.

PR #1534, opened 2021-02-08
Update hosts

GeriGeriGeriGeri:


funilrys:
@GeriGeriGeriGeri Thank you for your contribution and time. Can you please add a more detailed description to your PR ?

Thank you and stay safe and healthy.

StevenBlack:
@GeriGeriGeriGeri the file you changed is fetched from the `add.risk` source, at build time.  So the file you changed doesn't originate here.  This is all explained in this repo's [readme](https://github.com/StevenBlack/hosts/blob/master/readme.md).

Closing.

GeriGeriGeriGeri:
funilrys: i got a request from a friend to investigate an issue originating from his virtualbox virtual machine (clean install in debian). it turns out that virtualbox communicates with that server without permission from the user, the connection is being automatically opened and continously kept up. We dont know what the communication stream contains, it could be innocent upgrade watcher, or a spyware that sends every key press. I have added it to my hosts to test it, it does not seem to break anything. 

StevenBlack: sorry, i will submit the update to that file then.

dnmTX:
@GeriGeriGeriGeri as much as i appreciate your intentions,those domains,in their current state are either parked or doesn't resolve due to being incomplete.Therefore...useless,unless you do wildcard blocking which is not possible here.
This is the best description i found to describe the seriousness of them not being blocked:
**akamai :The largest MiTM in Internet history by intel orgs fronting 'CDNs'** (MiTM=man in the middle attack)
...but you need to provide the actual active (sub)domains and not the root once only.
This is what currently active `akamai` domain should look like: `a2-21-34-89.deploy.static.akamaitechnologies.com` 
And please note,not all of them are bad,so don't just start posting here whatever you can find from internet.

PR #1587, opened 2021-03-29
remove hbbtv.mediaset.net

scafroglia93:
this is for smartv app


StevenBlack:
Lorenzo @scafroglia93 the `/hosts` file, the file changed in this PR, is generated from data.  

In this case the data comes from my ad-hoc data file.

I see no evidence the domain `hbbtv.mediaset.net` is reformed.

Closing.

Issue #1563, opened 2021-03-05
CNAME-cloaked trackers


DavidCWGA:
AdGuard has published a list of CNAME-clocked trackers. (And explained what a CNAME clocked tracker is, and how they work.)

https://github.com/AdguardTeam/cname-trackers

I'm not entirely sure of the format of their repo, I think the list is all the .txt files in this directory concatenated together, maybe:
https://github.com/AdguardTeam/cname-trackers/tree/master/trackers

It would be nice to see these ingested into the list.

dnmTX:
**ping @ameshkov** 
Any chance of implementing combine(curated of course) list in a hosts format ( aka `0.0.0.0 domain.com` ) 
or just a plain list without the not needed characters that are used in adblocking software so can be added here as a addition 
to enhance the protection of everyone who uses the lists here?
Thanks as allways 👍

ameshkov:
Hm, I guess we can do a plain list as well. Could you please file a feature request to the repo so that we didn't forget? :)

dnmTX:
@ameshkov DONE! 
**for refference see:** https://github.com/AdguardTeam/cname-trackers/issues/3
Will wait for update. Thank you and @AdguardTeam for the excellent work as allways 👍

durablenapkin:
Just wanted to comment and say that @AdguardTeam has generated combined lists: https://github.com/AdguardTeam/cname-trackers#the-lists

The list that would be of interest is: https://raw.githubusercontent.com/AdguardTeam/cname-trackers/master/combined_disguised_trackers_justdomains.txt

DavidCWGA:
I manually added the new AdGuard list to my blocker about a week ago and so far nothing new seems to be broken that wasn't broken already.

dnmTX:
Patience everyone,[it's not automated,yet](https://github.com/AdguardTeam/cname-trackers/issues/3#issuecomment-798907962)
We'll wait for official word from the @AdguardTeam 👍

dnmTX:
Steve @StevenBlack this one is ready for your review and approval 👍 
**Raw Link:** https://raw.githubusercontent.com/AdguardTeam/cname-trackers/master/combined_disguised_trackers_justdomains.txt

StevenBlack:
Thanks Dan @dnmTX and David @DavidCWGA!  👍 Let's have a look.

[...]

PR #1588, opened 2021-03-31
Introduction of the support of RAW lines/sources

funilrys:
This patch fixes https://github.com/StevenBlack/hosts/issues/1563#issuecomment-810688754.

Indeed, before this patch, the updater was not supporting a RAW list of domains as input.

Changes:

    matches_exclusions():
        (new) Support for rule formatted as 'example.com' along with the pre-existing '0.0.0.0 example.com'.

    normalize_rule():
        (edit) Apply DRY.

        (new) Support the normalization of the rule formatted as  'example.com' along with the pre-existing '0.0.0.0 example.com'.

    strip_rule():
        (new) Complete rewrite in order to strip all possible lines.

    *.py (files):
         Apply Black & isort.

    setup.cfg:
         Add black recommended/related settings.

Stay safe and healthy.
Nissar

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,2 +1,6 @@
 [flake8]
 max-line-length = 120
+extend-ignore = E203
+
+[isort]
+profile = black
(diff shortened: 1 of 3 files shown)


funilrys:
Oh well, it's the power of OSS 🎉 😜 

It was fun and actually a quick-to-write patch 😉. 

You're welcome ☺️!

PR #1604, opened 2021-04-08
paledega source added

paledega:
I detect some spywares and wrote into my hosts file.

PR #1605, opened 2021-04-08
paledega source added

paledega:
This pull request is fix of https://github.com/StevenBlack/hosts/pull/1604

I split some address. 

If you have any problem I can move to other list

### DIFF
--- a/data/paledega/update.json
+++ b/data/paledega/update.json
@@ -1,0 +1,9 @@
+{
+    "name": "paledega",
+    "description": "Additional trackers, spywares",
+    "homeurl": "https://gist.github.com/paledega/14039224debc2fd954b835cc2288ce57",
+    "frequency": "occasionally",
+    "issues": "https://gist.github.com/paledega/14039224debc2fd954b835cc2288ce57",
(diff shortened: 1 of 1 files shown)


paledega:
https://gist.github.com/paledega/14039224debc2fd954b835cc2288ce57

dnmTX:
Steve @StevenBlack just for the record i'm **ALL AGAINST** merging this !!!!
To say that the list is too aggressive will be understatement  👎 👎 👎

paledega:
I can create more splited hosts. I problem for me. 

also I think this adresses are not agressive.

StevenBlack:
I'm sorry @paledega I'm not merging a GitHub Gist that's 4-days old.

We only carry data from sources with established track records, with a verifiable history of curation and activity.  That's just how we roll, here.

Remind me in a year, and then we'll see.

Also we've never carried a Gist.  It's an interesting idea.  But how do people raise and track issues with a Gist?  An endless thread at the bottom of the Gist?  How is that supposed to work in practice?

Closing.

StevenBlack:
Look, @paledega, I really don't want to be unkind, but seriously now.   C'mon!

PR #1616, opened 2021-04-16
Added partners subdomain for camsoda

william81fr:
This is a fixed re-do of:
https://github.com/StevenBlack/hosts/pull/1615

So sorry about that

### DIFF
--- a/data/StevenBlack/hosts
+++ b/data/StevenBlack/hosts
@@ -339,6 +339,7 @@
 0.0.0.0 www.applog.camera360.com
 0.0.0.0 camsoda.com
 0.0.0.0 www.camsoda.com
+0.0.0.0 partners.camsoda.com
 0.0.0.0 camteens.to
 0.0.0.0 candygirls.cc
 0.0.0.0 canuck-method.com


StevenBlack:
Thanks William @william81fr.

Our goal here is not to remove companies from the internet.

Our goal is to stop their vectors for the distribution of malware, crapware, tracking, and other sortid tactics.

So I cannot accept this unless `partners.camsoda.com` is a specific vector for what we block.  This has not been demonstrated.

william81fr:
Of course you don't want to block the whole internet, that wasn't my assumption.
How would you feel about moving their domain to the porn list then, instead of the base one? Because currently, you are indeed blocking them (unless the user works around that, but I wouldn't expect most users to be aware of the technique).

dnmTX:
@william81fr `*.camsoda.com` was added here in 2018 by @StevenBlack: 8596b4792220f4cd167510fe2147fe7fec4eccfd very likely
for pushing advertisements or even worst,popus to unrelated to porn websites.
Your PR here belongs to the porn extension(list) and not to the main list unless it's justified,aka provide us with proof(link) that the domain engaging in such a behaviour mentioned above.
That said,mine as well report or make a PR to the [proper repo](https://github.com/Sinfonietta/hostfiles/issues) instead of keep pushing it here,which will get you to nowhere.

StevenBlack:
Thanks William @william81fr and thanks Dan @dnmTX for the assist.

Closing.

PR #1632, opened 2021-04-30
Remove dead link

PromoFaux:
Dead link - this is no longer an officially endorsed product


StevenBlack:
Hi Adam @PromoFaux thanks, good catch.

Can you please amend this as a change to readme_template.md?

You see this repo has 16 different home pages — drill down into `alternates/` and see there are readme pages for different mixes and variants, regenerated at each release.

PromoFaux:
For recursion, see #1633

PR #1633, opened 2021-05-01
Remove dead link (alt to #1632)

PromoFaux:
As requested in #1632 

Unsure if you wanted it in addition to, or instead of - figured it was easier to do as two separate patches - and then you can use one or the other or both or none :)


StevenBlack:
Thanks Adam @PromoFaux.  Merging!

Issue #1661, opened 2021-05-22
hoping you could add these...


piengeng:
0.0.0.0 xvtelink.com # ads with redirects
0.0.0.0 invol.co # tracking 
0.0.0.0 36c4.net # redirect to go.trafficrouter.io
0.0.0.0 go.trafficrouter.io
0.0.0.0 track.venatusmedia.com # tracking
0.0.0.0 amptrack.dailymail.co.uk # tracking
0.0.0.0 mailmetromedia.amp.permutive.com #ad
0.0.0.0 videos-cloudflare.jwpsrv.com # video ads
0.0.0.0 videos-fms.jwpsrv.com # video ads

StevenBlack:
Thank you Henry @piengeng.

Added in commit 7ad72d76.

Closing.

bnainar:
@StevenBlack, how do you verify these actually unwanted sites? Any tests? Just curious to this whole process

Thanks for making my life better

piengeng:
@bnainar I think there's too much for a single person to verify every single entries, I propose those because of my own testing, and personally added those to my pihole's blacklist and hoping others that uses this list could benefit from it. I sure hope the community could prove me wrong on those urls

StevenBlack:
Hi Nainar @bnainar,

Mostly I rely on curators who've been regularly maintaining their own lists for a long time. 

For ad-hoc additions like those in commit 7ad72d7 I typically check several things.

* Who's submitting the domains.  Is this person an anon egg on Github, or does this person have a verifiable history.
* The plausibility of the domains, including checking the name of the suggested subdomain, whether we list this root domain already, etc.  It puzzles me why some companies name their domains like `track.venatusmedia.com` (as above) which screams, "block me!".
* Often trusted helpers will analyze a suggested domain and comment before I've seen the suggestion.  This is nice since it lightens the load.

Often it's just instinct 😄

zchrykng:
Just ran into legit videos being blocked by `videos-fms.jwpsrv.com`

Not sure what your process is for reviewing things like this, but it is blocking legit videos in at least one app. I've added it to my white list for now, and maybe it is something I should address with the developers of the app, I assume they are unaware they are sharing a domain someone sharing video adds.

StevenBlack:
Hi Zachary @zchrykng thank you for your observation.

To clarify, it doesn't matter if legit videos — however that's defined — are blocked.  If the domain vehicles their users, or other scam behavior, then it's blocked.  

[...]

PR #1665, opened 2021-05-25
remove JW Player entries

SageBrushes:
The JW Player entries below are not used to serve ads, but video content on the JW platform. While it is possible for a video ad creative to be served by the JW platform,  it is not a major use-case of the JW platform  and should not be included in an ad blocker.

0.0.0.0 videos-cloudflare.jwpsrv.com  # video ads
0.0.0.0 videos-fms.jwpsrv.com  # video ads


llacb47 (approved):

StevenBlack:
See also #1661

StevenBlack:
Thank you @SageBrushes.

Merging.

PR #1676, opened 2021-06-03
Ran PyFunceble on «Steven Black's ad-hoc list»; ~42% of the domains are dead or down

DandelionSprout:
Essentially, I ran a beta version of https://github.com/funilrys/PyFunceble on the *hosts* file, and its output indicates that 1,338 of its domains are dead or unavailable.


dnmTX:
@DandelionSprout in order to determine which of the domains are really dead(many got put in a park mode the come back at some point) you need to run PyFunceble at least for 2-3 months and do some analitycs on your end.You can't just run it once and come to resolution.
For one thing i agree on those nonsense/gibberish domains,aka `568vovhs6ekbupjo.pro`. They all are short lived and replaced fast. 
Steve @StevenBlack will determine the outcome here,i just posting my opinion 👍

DandelionSprout:
I'll admit that temporary park modes are a potential issue that sadly(?) haven't been a topic in the rather small PyFunceble community so far, and also something I didn't know what as an actual thing that malware-site makers tended to do.

As a result of those things, I've tended to rely entirely single-runs of PyFunceble thus far. 😅

funilrys:
Hi there,

First of all, if you search (here) for the word "PyFunceble" or "Funceble" you should be able to find past discussions about this.

False-Positive is the thing I try to avoid the most while developing PyFunceble. It's not perfect yet, but it will one day be. It's one of the few reasons why people are not using it. 

Steve @StevenBlack knows the existence of [PyFunceble](https://github.com/funilrys/PyFunceble), he stars the project, and somehow supports me but chooses to not use it actively (outside of syntax check I guess). We had this discussion in the past and I have that feeling that this may come back again in the future as the "rather small PyFunceble community" will grow "in the dark"... 

One of the reasons why he chooses to limit it to syntax check (with a manual safety check) is to avoid false positives. He also thinks - which is not bad - that the trusted maintainers are the one who should handle this through their own strategy and expertise. 
I mean, most of those list comes from PR or searcher that take a lot of time to analyze a report (issue, email, discussion, article, news, etc.) or their network/honeypot. - I used to do that back when I had plenty of time 😄; now I try to automate a lot. 

PyFunceble is a great testing tool but can have the reverse effect on some subdomain with a short life. We should actually do more to avoid the possible false positive behavior of PyFunceble. But, it's heavy.

That's one of the reasons: 
* I wrote - back then - the inactive database (you already have one on your filesystem I guess) which is used (when active) to get the list of subjects to retest the next time you will test the same file again.
* PyFunceble does not test against 1 method but 4-6: SYNTAX, WHOIS, NSLOOKUP, (NETINFO - especially for local network), HTTP CODE, REPUTATION (if active).

All to reduce false positives.

There is one part that is missing right now on top of PyFunceble in my opinion: Sharing, Interconnection, and analytic (over several days) of - the status/data -  of a subject.

That's why I'm building (in my little free time) a little platform/WebAPI on top of PyFunceble which will be able to make the status from multiple trusted runners (CI, CLI, Web, ...) of PyFunceble available to any running instance - of PyFunceble - in this world. 
That's one way to reduce false positives. A domain that is active for a CLI runner in Germany and not in the US will be automatically flagged as ACTIVE for anyone (CLI, WEB, CI, ...) using the platform - for example. 
On the other hand, it will also do some testing of a submitted subject over time in order to provide a possible better accuracy based on the result of the last 30-60 days (dynamic over time).

It's not a one-day job, but it's an idea of mine for the future of the "niche" and small community that PyFunceble is. It will definitely start small as only a few people will use it but it has the potentiality to become bigger over time. 

A taste of one of the "microservice"  in charge of testing (only) can be found [here](https://github.com/PyFunceble/web-worker). It's way faster than the CLI of PyFunceble because there no such thing as "database", "auto continue" or other things that are there to make the CLI and testing experience (mostly of file) safe, smooth, and continuous (for those who do that). It's just a web server with a documented API... 

I may already know what Steve will respond to your PR but let's see what he thinks of it.

Thank you for taking the time to open this PR @DandelionSprout! 
Never give up, one day we will - eventually beat them (the bad one).

Always keep in mind rules 8, 20, 23, 35 of the [Gibbs's Rules](https://ncis.fandom.com/wiki/Gibbs%27s_Rules) 😉 

Stay safe and healthy, you all!

StevenBlack:
Thank you @DandelionSprout I like this PR.  I'm inclined to merge this.  Let me sleep on it.

My one concern is we have `speakol` domains at the very bottom of the list that Dan @dnmTX found just a few days ago.

I've run PyFuncible on my ad-hoc list before, and I like the haicut it gives.  

As mentioned above, I'm very wary of Type-II errors.  Malware that broadcasts into the void is potentially minutes away from broadcasting to a live domain, activated only when the time is right.  It's almost like the world needs a [SETI](https://www.seti.org/) project, but pointed at the internet rather than deep space.  I sense the US government's NSA and ECHELON do this already.  I wish they maintained a hosts block list.

dnmTX:
> My one concern is we have speakol domains at the very bottom of the list that Dan @dnmTX found just a few days ago.

Steve @StevenBlack i just checked them all,and yeah they don't have any A and AAA records but there is a IP assign to each of them and the registration doesn't expire till 2022. Likely they're not in use at the moment,but who knows in the feature.
My opinion,i think it's ok to remove them,if there is a any reemergence in the feature will add them back 👍

dnmTX:
...also add `0.0.0.0 live-cams-6.livejasmin.com` whenever you adding something else 👍

StevenBlack:
Merging.  Thank you everybidy.

PR #1709, opened 2021-07-13
Fix whitelisting issue when RAW list are given

### LINKED ISSUE #1687: Whitelist not working for minecraft-hosts and AdGuard CNAME
After a long time, I've pulled the last repo changes and I've realised `click.redditmail.com` is part of `AdGuard CNAME`, so I added it into my personal whitelist, but it's not working.
After taking a look, I realized not only this list but also `minecraft-hosts` aren't working when you add one of the domains to the whitelist and both are in the format:
```
thedomain.com
```
instead of the rest of lists that come as:
```
0.0.0.0 thedomain.com
```
or
```
127.0.0.1 thedomain.com
```
so I'm thinking that maybe it's a script bug, also considering that [both](https://github.com/StevenBlack/hosts/pull/1639) [lists](https://github.com/StevenBlack/hosts/issues/1563) are fairly new in the project and probably it's a format that wasn't considered to be whitelisted before?

funilrys:
# Facts

1. I overlooked the whitelisting of RAW / Plain list when I adjusted the code for the support of them.

2. This PR:

    - Fixes #1687.
    - Touches @alagos @dnmTX @avatartw.

# Changes

- `remove_dups_and_excl()` 

Inside the `remove_dups_and_excl()` function, we were matching against such lines:

```
0.0.0.0 example.org
```

and not 

```
example.org
```

This patch fixes the bug by adjusting the regex we use to process the whitelisting. That means that from now on, we are capable of whitelisting from both (hosts + RAW / Plain) formats.

- Blacking. Don't complain or discuss code formatting, just use [Black](http://github.com/psf/black).

### DIFF
--- a/testUpdateHostsFile.py
+++ b/testUpdateHostsFile.py
@@ -543,7 +543,9 @@ class TestGatherCustomExclusions(BaseStdout):
     # Can only test in the invalid domain case
     # because of the settings global variable.
     @mock.patch("updateHostsFile.input", side_effect=["foo", "no"])
-    @mock.patch("updateHostsFile.is_valid_user_provided_domain_format", return_value=False)
+    @mock.patch(
+        "updateHostsFile.is_valid_user_provided_domain_format", return_value=False
+    )
     def test_basic(self, *_):
         gather_custom_exclusions("foo", [])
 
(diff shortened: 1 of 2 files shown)


StevenBlack:
Thank you Nissar @funilrys! 🥳 🎉

Merging!

PR #1729, opened 2021-08-09
Updated Windows batch file to pass arguments to python script...

Warrentheo:
Updated Windows batch file to pass arguments to python script...  This enables extensions to work on Windows...

### DIFF
--- a/updateHostsWindows.bat
+++ b/updateHostsWindows.bat
@@ -49,7 +49,7 @@ if not exist "%WINDIR%\System32\drivers\etc\hosts.skel" (
 
 :UpdateHosts
 :: Update hosts file
-py updateHostsFile.py --auto --minimise
+py updateHostsFile.py --auto --minimise %*
 
 :: Copy over the new hosts file in-place
 copy /y /v hosts "%WINDIR%\System32\drivers\etc\"


funilrys (approved):
LGTM!

StevenBlack:
Thanks @Warrentheo!

Merging!

StevenBlack:
Thank you for the assist Nissar @funilrys.

Warrentheo:
Thank you both, love the tool, use it on as many machines as I can 😁

Warrentheo:
Turns out, enabling all but one of the extensions on Windows 10 brings my 12 core 24 thread 64gb laptop to its knees...  When rebooting, and logging in, the network stack gets so overloaded and borked, that the computer only has about a 30% chance of making it all the way onto the desktop, and shows errors about parts of the kernel/classes not being registered and all kinds of things fail, to the point that apparently some programs start corrupting themselves, requiring reloading Windows from backups...  Resetting the hosts file back to default managed to fix all these issues...

Not really a bug report for this program, since the issue is with the Windows DNS resolution system, just means that I will probably have to switch to something like PiHole instead... :-/

rautamiekka:
Did you use the compression option ?

Warrentheo:
On first attempt, that shortened the hosts file from around 100k lines to only around 13k, and first boot afterward seems to be working fine...  It took me a while to figure out what was causing my issues, since I had been using hosts for a while, and didn't have this issue for most of it...  I think it was partly due to the jump in line numbers from getting this working, and noticing that the DNS processor version of the svchost.exe file was taking up 95% cpu for 5 minutes after boot, which finally put me on the right track...

I will use the new compressed version to see if this issue returns...  Thank you for pointing me in the right direction :-)

StevenBlack:
Thanks for this @Warrentheo.

I'd be curious to know if other Windows users are seeing similar issues.

The size of our  hosts files has been creeping-up.  The full-extensions hosts file is 126,683 domains right now. A year ago we carried 83,978 domains.  So that's a 50% annual growth rate.

I sense we're probably too big, right now.

StevenBlack:
For the base hosts file:

* today: 92,495 domains
* one year ago: 56,634 domains

Growth of 35,861 domains, year-on-year, solely attributable to the base list.

Warrentheo:
Early report: So far the compression seems to have fixed the issues, since my system remains as quick as this overkill of a system should be...  

That said, the issues I was having were weirdly intermittent, and difficult to trace since it never showed up the exactly same way twice, and then compounded by the fact that as it went, it would corrupt things faster and faster to the point that Windows would no longer boot...  I would want to run for about 2 weeks or more to be confident in saying that it is fixed... 

You would think Microsoft would find a way to sort and store this list such that it should be negligible to search though, but I don't know enough about it to claim to know what I am talking about...

rautamiekka:
Yeah, I also can't see why M$ would keep using such an extremely old+inefficient method, especially if you already need to reboot the DNS Client anyway. Just load the file in RAM, using a random-access datatype they are already using, couldn't be THAT hard.

Warrentheo:
2 Weeks in and running flawlessly...  With the growth of the list, are we moving toward setting the '--compress' option by default?  I understand there is other software that doesn't like that, but how common is that?  Anyone else having issues with leaving it uncompressed?

StevenBlack:
Thank you for the suggestion @Warrentheo.

I don't think we'll be setting the `--compress` option as default anytime soon.

I sometimes think about the extent this hosts project is used and trusted downstream, both as a hosts file and as a source for derivative works.

In other words, userland for this project is vast.  Sometimes I see evidence that userland may be immense.

I really don't want to do anything that could mess-up users downstream.

Warrentheo:
Ah, I understand and agree with you on that one...  There is an in between option however, since it is is Windows that seems to be having this issue, what about changing the 'updateHostsWindows.bat' file that already has the --auto and --minimize options enabled to also include the --compress option...  This should target the fix to only the systems that would have an issue, without messing with the people on Linux or that use the python script instead...

`py updateHostsFile.py --auto --minimise --compress %*`

Don't really know enough to advocate either way, so I will leave it up to you...

Issue #1734, opened 2021-08-15
Block www.portuguesinhas.com [PORN]


jfoclpf:
Please block Block www.portuguesinhas.com 

safer to block everything under *.portuguesinhas.com because the name is porn suggestive 

It is clear PORN

Thanks

StevenBlack:
Thanks João @jfoclpf!

Ping @sinfonietta or @clefspeare13.

dnmTX:
> Ping @Sinfonietta or @Clefspeare13.

noticed who manages the repo ☝️ . If it was me i'd drop it in a heartbeat ❗

jfoclpf:
PR made here: https://github.com/Clefspeare13/pornhosts/pull/369

StevenBlack:
Thanks João @jfoclpf, closing.

PR #1735, opened 2021-08-15
issue #1734

jfoclpf:
porn site


dnmTX:
@jfoclpf  all the porn data is derived from those sources: https://github.com/StevenBlack/hosts/tree/master/extensions/porn
Choose either one,visit their repo,do pull request. Easy as that.

PR #1774, opened 2021-10-04
Drop Python 3.5 from CI and README.

XhmikosR:
It's EOL since December 2020.

We could also drop 3.6 since it's going EOL at the end of 2021. If you'd like that, let me know so that I update the PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,7 +10,6 @@ jobs:
       fail-fast: false
       matrix:
         python:
-          - 3.5
           - 3.6
           - 3.7
           - 3.8
--- a/readme_template.md
+++ b/readme_template.md
@@ -81,7 +81,7 @@ docker run -it (containerid) bash
 
 ### Option 2: Generate it in your own environment
 
-To generate your own amalgamated hosts files you will need Python 3.5 or later.
+To generate your own amalgamated hosts files you will need Python 3.6 or later.
 
 First, install the dependencies with:
 


StevenBlack:
Hi @XhmikosR!

I'm not aware of any reason for this. Is 3.5 holding something back?  I'm not saying no, just asking, why?  If 3.5 is not a blocker then I'm inclined to just leave it.

I should clarify, my background and experience leaves me with a profound respect for userland, and the installed base. Every time churn is forced upon me and my project mates for gratuitous reasons, I seethe inside. It really bothers me. I always swear, that's not me. I don't want to inflict pain without clear reasons, and even then, only after extended pain until I can't take it anymore.  That's when I fold, and agree to pass-on the pain.

Some committee somewhere decided 3.5 is EOL. That's fine, that's their prerogative.  We aren't doing anything fancy here.  We can probably support 3.5 for a long, long time, with zero pain.

Or am I wrong about that?

rautamiekka:
My 2c:

When you put it like that, it certainly doesn't change anything switching to 3.6 since no code changes are made, UNLESS CI fails to use 3.5, which I doubt will be a problem anytime soon.

PyCharm and testers will warn when some new feature can't be used in an earlier Python.

funilrys:
Actually, I don't see any reason to officially deprecate it here. The scripts were written to support it. New features of Python 3.6 are explicitly not used to allow retro compatibility. 

I always assume that for a repository as large as this one, we can't guarantee that the scripts are not constantly downloaded by old machines with an old version of Python.
Reaching EOL for a language like Python and a "minor" version like 3.5 doesn't mean that it is not available any more on machines around the globe. 
I could use the latest feature of Python 3.8+ in this project but I have to think of the millions of machines that still have 3.5. Therefore, I discipline myself (in this scope) to ignore the nice and recent features that are available to me.


Another thing to keep in mind: The most used Linux distributions on earth still provide Python 3.5 as default. Next year maybe when the version of the distribution in questions reaches OES ... But not yet ...

I personally won't add or validate all the nice and shiny features of python 3.6+ just because we decide to "deprecate Python 3.5". Most people have their own environment, machine, and discipline. They should just know that we support everything from Python 3.5. 
Even if our CI fails, we should discipline ourselves until next year - at least. 

An example: Debian 9 is supported until June 2022. And in Debian 9 officially distribute Python 3.5.
It's just too early in my opinion. Most distributions are not that far yet.

Your decision Steven @StevenBlack. I personally won't add or validate all the nice and shiny features of python 3.6+. It should go through CI with Python 3.5 until at least the middle of next year.


P.S.: I'm aware that people may use `conda` or other kinds of Python Virtual Environment. Those people are not taken into consideration in my comment. For them, everything will just run.

StevenBlack:
Thank you Nissar @funilrys for these insightful comments.

Thank you @XhmikosR for the thoughtful PR but I've decided to decline this.  Please don't let this deter you from future PRs 😄

PR #1775, opened 2021-10-04
A few npm updates

XhmikosR:
@StevenBlack: is `readmeData.json` supposed to be committed? I'm asking because I see it's in .gitignore: https://github.com/StevenBlack/hosts/blob/a3bdc8b71b5577b252e2df018966d084a18a0545/.gitignore#L6

BTW I assume you don't want the package-lock.json file committed hence the second patch. Otherwise let me know and I'll adapt the PR :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,4 +7,5 @@ readmeData.json
 hosts-*
 /web.config
 /__pycache__
-/node_modules
+/node_modules/
+/package-lock.json
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,1 @@
+package-lock=false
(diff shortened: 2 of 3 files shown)


funilrys (approved):
Looks good to me!
It's yours @StevenBlack!

StevenBlack:
Thanks @XhmikosR!  Merging.

XhmikosR:
@StevenBlack how about `readmeData.json`? See my OP :)

StevenBlack:
@XhmikosR yes, thank you, I realize that.  It's wrong, yes, but it's not hurting anything. So it's semantically wrong, but with no side effects.  Fixed.

PR #1776, opened 2021-10-05
Add Nix Flake

Deleh:
[Nix Flakes](https://nixos.wiki/wiki/Flakes) are an upcoming feature of the Nix package manager. They allow truly reproducible builds and provide an standardized interface to Nix things like packages, development shells and NixOS modules.

This flake provides a development shell and a NixOS module.
Every user which has a recent version of Nix (and Flakes are enabled) can start a development shell with all requirements fulfilled like this:

```
nix develop
```

NixOS users which use a flaked configuration can use the hosts module in the following way:

```
inputs.hosts.url = github:StevenBlack/hosts;
outputs = { self, nixpkgs, hosts }: {
  ...
  modules = {
    hosts.nixosModule {
      networking.stevenBlackHosts.enable = true;
    }
  };
  ...
};
```

I've also added four options for the host extensions. They can be used like this:

```
networking.stevenBlackHosts = {
  enable = true;
  blockFakenews = true;
  blockGambling = true;
  blockPorn = true;
  blockSocial = true;
};
```

If you like this I can update the README as well :)

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,40 @@
+{
+  "nodes": {
+    "flake-utils": {
+      "locked": {
+        "lastModified": 1631561581,
+        "narHash": "sha256-3VQMV5zvxaVLvqqUrNz3iJelLw30mIVSfZmAaauM3dA=",
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "rev": "7e5bf3925f6fbdfaf50a2a7ca0be2879c4261d19",
+        "type": "github"
+      },
+      "original": {
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "type": "github"
+      }
+    },
+    "nixpkgs": {
+      "locked": {
+        "lastModified": 1633267966,
+        "narHash": "sha256-gFKvZ5AmV/dDTKXVxacPbXe4R0BsFpwtVaQxuIm2nnk=",
+        "path": "/nix/store/k13ripsl4n2p6wf2ksy5m017ryykx4qc-source",
+        "rev": "7daf35532d2d8bf5e6f7f962e6cd13a66d01a71d",
+        "type": "path"
+      },
+      "original": {
+        "id": "nixpkgs",
+        "type": "indirect"
+      }
+    },
+    "root": {
+      "inputs": {
+        "flake-utils": "flake-utils",
+        "nixpkgs": "nixpkgs"
+      }
+    }
+  },
+  "root": "root",
+  "version": 7
+}
--- a/flake.nix
+++ b/flake.nix
@@ -1,0 +1,45 @@
+{
+  description = "Unified hosts file with base extensions.";
+  inputs.flake-utils.url = "github:numtide/flake-utils";
+  outputs = { self, nixpkgs, flake-utils }: {
+    nixosModule = { config, ... }:
+      with nixpkgs.lib;
+      let
+        cfg = config.networking.stevenBlackHosts;
+        alternatesList = (if cfg.blockFakenews then [ "fakenews" ] else []) ++
+                         (if cfg.blockGambling then [ "gambling" ] else []) ++
+                         (if cfg.blockPorn then [ "porn" ] else []) ++
+                         (if cfg.blockSocial then [ "social" ] else []);
+        alternatesPath = "alternates/" + builtins.concatStringsSep "-" alternatesList + "/";
+      in
+      {
+        options.networking.stevenBlackHosts = {
+          enable = mkEnableOption "Use Steven Black's hosts file as extra hosts.";
+          blockFakenews = mkEnableOption "Additionally block fakenews hosts.";
+          blockGambling = mkEnableOption "Additionally block gambling hosts.";
+          blockPorn = mkEnableOption "Additionally block porn hosts.";
+          blockSocial = mkEnableOption "Additionally block social hosts.";
+        };
+        config = mkIf cfg.enable {
+          networking.extraHosts =
+            builtins.readFile (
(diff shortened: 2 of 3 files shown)


funilrys (commented):
Should we really maintain this for so "few" people? I don't see a necessity here @StevenBlack.

Lock file on this scale may not be a good idea ... People who develop this project are free to have their development environment.

 We just require that everything that goes through the CI check and is formatted with Black. Any other development-related configuration that is not explicitly given all those years is just something that Steve, me, or any other active developer has to maintain.   

The feature is good for users of NixOS but not necessary for general development/usage.
I just see this as something else to maintain Steve @StevenBlack.

funilrys:
It's yours Steve @StevenBlack!

StevenBlack:
Thanks for your input Nissar @funilrys.

I think this is cool, actually.  I don't use NixOS but we already provide special goodies for Windows users, for example, and special instructions for various flavors of Linux.  I don't see the harm in this and, if we ever have issues, we can ping @Deleh 😄 

@Deleh before I merge this, can you please... 

- [ ] document this in the [`readme_template.md`](https://github.com/StevenBlack/hosts/blob/master/readme_template.md) file at the bottom of the [Linux section](https://github.com/StevenBlack/hosts/blob/master/readme_template.md#linux).  Once we have that then we're probably good-to-merge.

funilrys:
Then, flake is unnecessary. It's only useful for development not for runtime/installation.

StevenBlack:
Am I missing something Nissar @funilrys?  In my admittedly rudimentary understanding, is this not build sugar?

funilrys:
Well @StevenBlack, from my understanding .... if we take this as a build/installation process,

Then, the following lines:

```
[...]
inputs.flake-utils.url = "github:numtide/flake-utils";
[...]
python3Packages.flake8
```

are unnecessary as flake8 is there just for checking the source code on development. 
I assumed this PR is "development environment" related because of this statement:

```
Every user which has a recent version of Nix (and Flakes are enabled) can start a development shell with all requirements fulfilled like this:
```

So for me, this wasn't built for installation but really for a development environment. 

But if we now want to provide/integrate this as an installation/configuration provider - which `nixosModule = { config, ... }: .....` do... It should just provide the thing that is really necessary for installation/configuration. `flake8` shouldn't be part of the installation/configuration as it is part of the development toolchain (and it's really restricted to those like me who make some modifications to the python scripts). 

Also, Steve @StevenBlack, let's avoid lock files!!!! Lock files force us to maintain them over time. And I'm sure none of us have the time to constantly maintain a lock file ... Pinging someone to update lock files is never a good idea...

Deleh:
Thank you @funilrys and @StevenBlack for the review.

The development shell was only added for completeness, it can easily be removed again. Then the mentioned part would be gone:
```
[...]
inputs.flake-utils.url = "github:numtide/flake-utils";
[...]
python3Packages.flake8
```
The useful thing of the flake is actually the module, so every NixOS user can install the hosts file on his system with the small configuration snippet.
I would be happy to update the lock file every now and then, so there's no need for pinging. And if there should occur some issues I will absolutely fix them :)

Deleh:
> @Deleh before I merge this, can you please...
> 
> * [ ]  document this in the [`readme_template.md`](https://github.com/StevenBlack/hosts/blob/master/readme_template.md) file at the bottom of the [Linux section](https://github.com/StevenBlack/hosts/blob/master/readme_template.md#linux).  Once we have that then we're probably good-to-merge.

@StevenBlack Shall I extend the [Using NixOS](https://github.com/StevenBlack/hosts/blob/master/readme_template.md#using-nixos) section in the readme, too?

Deleh:
> @Deleh Yes I think that's where it should go. Good catch. The placement of the NixOS section seems wrong to me. I leave you to assess and revise how the Linux story should play out in the readme_template file.

Okay thank you :) I will submit the changes soon.

> Nissar @funilrys I'm cool with this particular lock file. This isn't a very dynamic `.lock` file like many others we are familiar with. Is that presumption correct @Deleh?

@StevenBlack this is correct, the lockfile just pins to a state when it was 100% working and makes sure that every other user gets exactly the same working state on his system. There should actually never be any need to update it but I will take care to keep it fresh if this is okay for you.
And I could still throw out the development shell to drop the dependency on `numtide/flake-utils`. There are so few package dependencies that a Nix development shell can easily be launched in a simple oneliner:
```
nix-shell -p python3 python3Packages.flake8 python3Packages.requests
```
So there is really no need for it.

StevenBlack:
@Deleh are there any NixOS-specific instructions for the [Reloading hosts file](https://github.com/StevenBlack/hosts/blob/471a2d41dddb12967a95c23368b60a61e05c33a7/readme_template.md#reloading-hosts-file) section? We mention many Linux distributions there, but not NixOS.  What do you think?

Deleh:
@StevenBlack I've added a section for completeness but there are no user actions required. Nix cares about reloading the `nscd.service` when the hosts get changed.

StevenBlack:
@Deleh yeah I realize now, NixOS is a package manager, not a fringe Linux distribution. I actually tried to find a Docker operating system image for NixOS thinking, I should check that out, maybe it's awesome 🙄 😆 

Merging!

Deleh:
@StevenBlack This is great, thank you very much :)

toastal:
> And I could still throw out the development shell to drop the dependency on `numtide/flake-utils`

You can drop the `flake-utils` dependency & still keep a development shell (tho the preferred API has changed a little since this was opened from `{ outputs.devShell = … }` to now `{ outputs.devShells.default = … }`). There is no need for `flake-utils` just to call a loop (see [2421](https://github.com/StevenBlack/hosts/pull/2421)).

PR #1827, opened 2021-12-12
Update CI config

XhmikosR:
* `actions/setup-python` now has built-in support for caching
* add `workflow_dispatch` to add the ability to manually trigger the workflow
* add Python 3.10

I still think we should remove the no longer updated Python 3.5 from CI; now we test 6 Python versions times 3 = 18 builds. Plus there are some warnings on macOS for Python 3.5: https://github.com/StevenBlack/hosts/runs/4498028405?check_suite_focus=true#step:3:1

Alternatively, we could adapt the CI matrix to test on all Python versions only for Ubuntu, and test only 1-2 on the other platforms.

@StevenBlack Either way, let me know and I can make any further changes :)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,20 +1,25 @@
 name: CI
 
-on: [push, pull_request]
+on:
+  push:
+  pull_request:
+  workflow_dispatch:
 
 jobs:
   test:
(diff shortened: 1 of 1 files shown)


StevenBlack:
Thank you @XhmikosR for keeping this facet of the repo current! 👍🏻 

Merging.

PR #1828, opened 2021-12-12
Update CodeQL workflow

XhmikosR:
* bring in the latest upstream changes
* add `workflow_dispatch` so that the workflow can be triggered manually
* remove autobuild since we don't need to build anything
* reindent

Non-whitespace diff: https://github.com/StevenBlack/hosts/pull/1828/files?w=1


StevenBlack:
Thanks @XhmikosR!  Merging.

Issue #1830, opened 2021-12-12
Feature Request: dnscrypt-proxy support


lsm5:
AFAICT, dnscrypt-proxy can't use the hosts file as-is so I end up running a few seds on the hosts file.
Would you be open to adding a version readily consumable by dnscrypt-proxy?

StevenBlack:
Hi Lokesh @lsm5 thank you for this suggestion.  

Can you share your `sed` commands so I can assess?  What format does `dnscrypt-proxy` require? Is it domain only, with a terminating period? One per line or is space-delimited okay too?

lsm5:
@StevenBlack thanks for the quick response and really appreciate your work.. And my bad, I wasn't aware dnscrypt already has a tool for this https://github.com/DNSCrypt/dnscrypt-proxy/wiki/Combining-Blocklists .

I guess that tool is good enough for me and this issue can be closed. But I'll leave it open in case you wanna add it here, maybe in the form of a docs update.

lsm5:
I'll be happy to send a PR if it's worth adding to the hosts repo.

StevenBlack:
Lokesh @lsm5 I'm always happy to improve our docs.

If you make improvements to docs, they should go in [the readme_template](https://github.com/StevenBlack/hosts/blob/master/readme_template.md). This repo has 17 readme files, including the root readme, all of which are generated from the template.

lsm5:
@StevenBlack ack, thanks a lot! I'll go through it and send something maybe within the next couple of days.

dnmTX:
Steve @StevenBlack this what they use in general:
```
 ######################################################
 #        Pattern-based blocking (blocklists)        #
 ######################################################
 
 ## Blocklists are made of one pattern per line. Example of valid patterns:
 ##
 ##   example.com
 ##   =example.com
 ##   *sex*
 ##   ads.*
 ##   ads*.example.*
 ##   ads*.example[0-9]*.com
 ##
 ## Example blocklist files can be found at https://download.dnscrypt.info/blocklists/
 ## A script to build blocklists from public feeds can be found in the
 ## `utils/generate-domains-blocklists` directory of the dnscrypt-proxy source code.
```
Looks like the very basic,without any wildcard blocking is just plain domains,no zeroes in the front 👍

PR #1833, opened 2021-12-13
docs: Include link to dnscrypt-proxy blocklist tool

lsm5:
Copy-pasted the tool description from the dnscrypt-proxy page.

Fixes: #1830

@StevenBlack @dnmTX PTAL . I think that should suffice, but let me know. Thanks!


StevenBlack:
Thank you Lokesh!  Merging.

PR #1852, opened 2022-01-02
mypornsnap.top

FadeMind:
ESET blocked it results in this search engine site against JS/ScrInject.B trojan.


StevenBlack:
Thank you Tomasz @FadeMind.  Merging!

PR #1892, opened 2022-02-09
remove domains PyFunceble mark as "INACTIVE"

MrRawes:


StevenBlack:
Thank you @MrRawes.  Merging.

PR #1911, opened 2022-03-06
CI: update actions

XhmikosR:
* actions/checkout to v3
* actions/setup-python to v3

@StevenBlack:

1. I still think we should drop Python 3.5, at least from CI since it longer works and make the whole build show as failed :/
2. Shouldn't we run `python3 makeHosts.py` on CI to make sure the scripts work?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,7 +33,6 @@ jobs:
         uses: actions/setup-python@v3
         with:
           python-version: ${{ matrix.python }}
-          cache: pip
 
       - run: python --version
       - run: pip --version
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,8 +34,10 @@ jobs:
         with:
           python-version: ${{ matrix.python }}
 
-      - run: python --version
-      - run: pip --version
+      - name: Print versions
+        run: |
+          python --version
+          pip --version
 
       - name: Install Python dependencies
         run: pip install -r requirements.txt
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,10 +27,10 @@ jobs:
 
     steps:
       - name: Clone repository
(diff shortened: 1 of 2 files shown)


funilrys:
Our policy was - and is probably still is: Every functions should be tested. Therefore, there is no need to run the script in the CI/CD workflow. 
I had to learn it the hard way in the past (thanks @gfyoung!) but I have to agree in that one thing: It give us confidence that it's running. 

About the 3.5 error. Can you try to disable pip cache to see how it reacts ? I'm just wondering ... I have private workflows running without any issues with python 3.5 ...

XhmikosR:
I removed pip cache and it seems to fix the issue. That being said, I strongly think we shouldn't test/support Python 3.5 at some point. Setting up Python 3.5 seems to be one of the slowest workflows.

As for running the actual scripts, I'll leave this to you guys. We could probably automate releases too later.

.github/workflows/ci.yml
```diff
@@ -27,16 +27,17 @@ jobs:
...
       - name: Set up Python
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v3
         with:
           python-version: ${{ matrix.python }}
-          cache: pip
 
-      - run: python --version
-      - run: pip --version
+      - name: Print versions
```
XhmikosR, line 37: This is also slow on macOS and Windows. We can already see the Python version, so maybe we should drop this step?

StevenBlack:
Thank you @XhmikosR! This was literally the next thing on my list and I'm glad you're addressing this right now.

I'd like to defer these decisions you your better judgment because the finer points of Python testings is not my forte.

I'm away from the office on a little break, so here's my plan: I'm going to iterate your PRs in chronological order, and then I'm going skiing :-)

Thank you for addressing this. You're the best.

XhmikosR:
Have fun! I'll try to think about the remaining issues later and see what I come with :)

PR #1912, opened 2022-03-06
Rename disable-dnscache-service-win to bat

XhmikosR:
Now it's consistent with the other bat file and .editorconfig/.gitattributes apply to this file too.


funilrys (commented):
This may be a breaking change for the people who automatically download the "latest" `cmd` file - without git.

XhmikosR:
Yeah, maybe, but it was wrong. It was using LF line endings.

XhmikosR:
To see why I have already explicitly set up CRLF in our .gitattributes/.editorconfig: https://serverfault.com/questions/429594/is-it-safe-to-write-batch-files-with-unix-line-endings

So, we could add `.cmd` too but IMHO this is simpler and more consistent.

StevenBlack:
I'm going to merge this now. Nissar's @funilrys comment is a good one, I think the remediation path for that appears straightforward, I hope that's true 🙃

XhmikosR:
The thing is that Windows cmd can fail with LF (with cryptic error messages) and this was the fix with the less changes.

PR #1916, opened 2022-03-07
Use `assertIn` instead of `assertTrue`

XhmikosR:
Fixes one of the two issues from https://lgtm.com/projects/g/StevenBlack/hosts/latest/files/testUpdateHostsFile.py?sort=name&dir=ASC&mode=heatmap&showExcluded=true

I'll leave the second on to someone else (if it really needs to be fixed).

I wonder what this line is supposed to do? https://github.com/StevenBlack/hosts/blob/e6da3b0fbfa194e6532247fbbc2b0ebde0fd6dc9/testUpdateHostsFile.py#L57


StevenBlack:
I learn something new every day.  Thanks @XhmikosR!

Merging.

PR #1919, opened 2022-03-12
Update CI

XhmikosR:
* drop Python 3.5
* add pip caching back
* run makeHosts.py

Notes:

- fixes CI errors/warnings:
  image
- reduces CI time approx. to half (**4:30 -> 2:30**)
- I added a `makeHosts.py` run too, we can drop this if you don't like it, but this gives us a chance to automate releases and also test things as close as possible to local
- I'd still drop Python 3.5 support from docs/script itself so that we are consistent and so that we can update deps, see #1920

### DIFF
--- a/readme_template.md
+++ b/readme_template.md
@@ -81,7 +81,7 @@ docker run -it (containerid) bash
 
 ### Option 2: Generate it in your own environment
 
-To generate your own amalgamated hosts files you will need Python 3.5 or later.
+To generate your own amalgamated hosts files you will need Python 3.6 or later.
 
 First, install the dependencies with:
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,6 @@ jobs:
       fail-fast: false
       matrix:
         python:
-          - '3.5'
           - '3.6'
           - '3.7'
           - '3.8'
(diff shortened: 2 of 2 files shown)


XhmikosR:
@StevenBlack here is the PR, please check the comments above and let me know :)

funilrys (approved):
LGTM!

readme_template.md
```diff
@@ -81,7 +81,7 @@ docker run -it (containerid) bash
 
 ### Option 2: Generate it in your own environment
 
-To generate your own amalgamated hosts files you will need Python 3.5 or later.
+To generate your own amalgamated hosts files you will need Python 3.6 or later.
```
funilrys, line 84: This is actually a lie. It can run with Python 3.5. We are just not supporting (as in giving support/help) it anymore.

XhmikosR: Please, I'm testing things...

XhmikosR: And to answer, no you won't be able to use Python 3.5 after #1920.

funilrys (approved):

StevenBlack:
Thank you @XhmikosR and Nissar @funilrys!

Open source: where software evolves while you sleep 😄.

Merging!

PR #1920, opened 2022-03-12
Update requirements.txt

XhmikosR:
* limit requests to v2.0
* require flake8 v4.0

Should be merged after #1919 due to Python 3.5.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,2 @@
-requests>=2.20
-flake8>=3.8,<=4.0
+requests ~=2.20
+flake8 ~=4.0


funilrys (changes requested):
Flake can stay permissive. It's the tool that let us know something is wrong. Therefore it is recommended that it evolves. It is actually a test dependency. 

Keeping it up to date let us know when we did something wrong in the past.

XhmikosR:
Well, not limiting deps is not a good practice. We can have dependabot update it if there's a major version.

XhmikosR:
#1921

funilrys (approved):

requirements.txt
```diff
@@ -1,2 +1,2 @@
-requests>=2.20
-flake8>=3.8,<=4.0
+requests>=2.20,<3.0
```
XhmikosR, line 1: @funilrys since you are more familiar with Python, wouldn't this also work? It seems smaller and matches what people use on npm.

```
requests ~=2.20
flake8 ~=4.0
```

Not sure if it makes sense in the pip world, though.

funilrys: Hey @XhmikosR, yes that will work. It will restrict to the major version.

XhmikosR: Thanks! I'm gonna switch to that and we can revisit if needed.

StevenBlack:
Thank you @XhmikosR and Nissar @funilrys!

Merging.

PR #1921, opened 2022-03-12
Add dependabot for pip and action updates

XhmikosR:
Happy to adapt the values, I went with sane values, but you we can change the frequency.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,17 @@
+version: 2
+updates:
+  - package-ecosystem: "pip"
+    directory: "/"
+    schedule:
+      interval: weekly
+    labels:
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,17 @@
+version: 2
+updates:
+  - package-ecosystem: "pip"
+    directory: "/"
+    schedule:
+      interval: weekly
+    labels:
+      - dependencies
```
XhmikosR, line 8: These labels will need to be added to the repo

StevenBlack: Labels added!

StevenBlack:
Thank you @XhmikosR!

Merging!

PR #1923, opened 2022-03-12
Update dependabot.yml

XhmikosR:
@StevenBlack Sorry about that, I'm not used to `pip` and made a bad copy/paste there :)


StevenBlack:
Thanks @XhmikosR! Merging this now. This has no effect on the latest release, so all-good!

Issue #283, opened 2017-04-29
New source: Anti popads.net


betterwebleon:
Hello,
I've found a new data source on the website [FilterLists](https://filterlists.com/), where it scores highly.

It is called "[Anti popads.net](https://github.com/Yhonay/antipopads)". Raw file is [here](https://raw.githubusercontent.com/Yhonay/antipopads/master/popads.txt). Currently it has more than 970 entries and is updated frequently. Description: _"Blocks shady, annoying pop-under ads from the infamous PopAds ad network."_

Could you add it among sources for your unified hosts file?

StevenBlack:
@betterwebleon sorry for the delay.  Hey thanks for pointing me to [FilterLists](https://filterlists.com/); that's cool!  I like how they present the sources and their contact facets.

I'm Currently traveling.  I plan to revisit this very soon.  Thanks for the suggestion.

betterwebleon:
No problem, there's no hurry. Thank you for considering it!

FilterLists is comprehensive and well structured website indeed. I found it just recently, thanks to @gorhill from his [wiki page](https://github.com/gorhill/uBlock/wiki/Filter-lists-from-around-the-web) about custom filter lists for [uBlock Origin](https://github.com/gorhill/uBlock/) (his content blocker).

Yhonay:
Hi, Yhonay here. I've added a list in a form of a hosts file to my git repo. I don't know how usefull it will be seeing how often popads.net adds new domains to their repertoire but if you'll find a use for it then [here it is](https://raw.githubusercontent.com/Yhonay/antipopads/master/hosts).

StevenBlack:
Hi @Yhonay!  How are you sourcing the domains in [antipopads](https://raw.githubusercontent.com/Yhonay/antipopads/master/hosts)?  What's your process to find these?

Yhonay:
Someone linked me to a list of a few thousands sites using popads.net. I pick a few dozen at random, unobfuscate everything between <script></script> and try to download any found .js file. Then I check if it matches a few keywords found in a standard popads.net .js file. If it matches then a domain that .js was downloaded from is added to the list.

I know it's a little naive but so far it works.

FadeMind:
@StevenBlack  @betterwebleon 
https://github.com/FadeMind/hosts.extras/commit/6ad79e7e2de999bea0eb26bf9bf69115e074c1d2

XhmikosR:
[...]

PR #1944, opened 2022-04-11
add antipopads-re

FadeMind:
Add `Antipopads:reborn` aka `antipopads-re` repo. 
https://github.com/AdroitAdorKhan/antipopads-re
Reference https://github.com/StevenBlack/hosts/issues/283#issuecomment-1095066066


StevenBlack:
Thank you Tomasz @FadeMind but I've decided to decline.

See the discussion in #283.

In short: this would make the list 24% larger, make it minimally more effective, while causing problems with less capable implementations, like Windows OS for example.

Closing.

PR #1991, opened 2022-06-28
clevertap

ahsandar:
Cleavertap is a tracking platforma that tracks users by across devices and services by  fingerperinting them


StevenBlack:
Thank you Ahsan @ahsandar!

Merging.

PR #1999, opened 2022-07-05
Container automation

rhtenhove:
This makes it possible to, and provides an example on, how to run the whole `/etc/hosts` replacement using just a container.

This required various changes throughout the code, these are listed below:

1. Reduced the image size from ~1GB to ~140MB
2. Sets up container pushing to `ghcr.io` on pushes to `master` and tags so no need to build/git clone yourself
3. Made scripts executable and added to PATH
4. Removed image digest. Just introduces something to maintain, and will probably introduce more CVEs through being outdated, than preventing rogue registries. Since this PR also introduces using the `ghcr.io` image built here, the GHA builders need to be compromised, and at that point the problems are bigger than a digest can solve.

While I was at it, also introduced some fixes:
1. Only flush cache if the file move succeeded
2. Replace Windows hosts file from python script
3. Removed the EOL 3.6 from testing
4. Check for existence of source hosts file before trying to copy

Using it myself now, and works beautifully without needing anything besides Docker.


StevenBlack:
Hey @rhtenhove this is so interesting. ✨ Thank you for this! 

I'll be evaluating this shortly.

StevenBlack:
@rhtenhove this looks brilliant!

This beats my rudimentary and cursory knowledge of Docker.

rhtenhove:
Thanks! :smile: 

I do seem to have broken all the tests. Seems to mostly be changes in text output. I'll try to have a look at those later.

rhtenhove:
Should be good now

rhtenhove:
Now it builds on tags as well, as the docker meta action only makes it push to `latest` on a tag. This means it'll also produce images with versioned tags.

StevenBlack:
Hi @rhtenhove there are words there I understand 😄 

I like everything I see here, I just need to upgrade my Docker-foo to fully appreciate all this.  I should be able to do that soon. ⏲️

.github/workflows/ci.yml
```diff
@@ -17,7 +17,6 @@ jobs:
       fail-fast: false
       matrix:
         python:
-          - '3.6'
```
StevenBlack, line 20: 👍🏻

.dockerignore
```diff
@@ -0,0 +1,7 @@
+*
+!updateHostsFile.py
+!data
+!alternates
+!extensions
+!readmeData.json
+!requirements.txt
```
StevenBlack, line 7: @rhtenhove can you briefly explain why these files belong in `.dockerignore`?  This seems counter intuitive to me but, then again, what do I know?

.github/workflows/container.yml
```diff
@@ -0,0 +1,43 @@
...
+        uses: docker/metadata-action@v4
+        with:
+          images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
+      - name: Build and push container image
+        uses: docker/build-push-action@v3
+        with:
+          context: .
+          push: ${{ github.event_name != 'pull_request' }}
+          tags: ${{ steps.meta.outputs.tags }}
+          labels: ${{ steps.meta.outputs.labels }}
```
StevenBlack, line 43: @rhtenhove are these `env.` and `secrets.` elements going to be problematic for those who fork this repo?

Are they going to be problematic for me? 😄

StevenBlack:
@rhtenhove one additional question: why is file `updateHostsWindows.bat` an empty file now?

StevenBlack:
Ping @rhtenhove 😄

.dockerignore
rhtenhove: It's actually a reverse. So we dockerignore everything with `*`, except (`!`) for the files and folders we actually need. Allows for a much cleaner ignorefile, prevents adding unneccesary new stuff, and improves caching as not every little update busts the cache.

rhtenhove:
> Ping @rhtenhove smile

Sorry, missed your reviews! They went in the bin together with your first 'I'll take a look soon' comment in my e-mail client. Protonmail's phone client isn't very clear with their conversation view...

.github/workflows/container.yml
rhtenhove: The job only runs on `master` and on tags, i.e. people forking it correctly should not ever see this job run on their own fork.
For those that do push to their own `master`, will have the job run with their own credentials, pushing the image to their own registry. Also not a problem, because when they delete their fork after sending in a PR, the images will be gone as well.

The secret used here is an automatic one, made available to any GHA workflow with proper access restrictions, so only pushes done by you (or anyone else with full access to your repo) to `master` will actually build an image and push it to the ghcr.io/StevenBlack/hosts registry.

If anything goes wrong, the job will fail and no harm done, we can fix it later. For example, I'm unsure if it will complain about using capital letters in the repo name.

rhtenhove:
> @rhtenhove one additional question: why is file `updateHostsWindows.bat` an empty file now?

It's not, it just didn't have any content changes (I do agree that that is a terrible message to display there). The only change is the `mode`, i.e. it is now executable. Although that doesn't mean much in Windows contexts.

.github/workflows/container.yml
StevenBlack: Thank you for this clarification @rhtenhove.

I'm wary of introducing dependencies, especially environment variables.

Also, small point: making Pull Requests is one of many reasons people fork a Github repository, and the presumption that folks delete forks once PRs are done doesn't hold, at least in my experience.

rhtenhove: The `env` key is introduced in the file itself (at the top) and is about the environment of only the job in that file. The secret GITHUB_TOKEN is also a convenient token, as it ensures it'll behave the same for everyone, without having to set anything up.

Finally, forking and then deleting or not doesn't matter, so my presumption is actually irrelevant. It is meant to clarify that it will not have _any_ consequences for people forking, and nobody will ever need to do anything with the content or results of that job if they don't wish to.

That's so great about CI and the provided default variables :-)

StevenBlack:
Thank you @rhtenhove this is one hundred percent awesome. 👍🏻 

Merging!

PR #2008, opened 2022-07-13
readme fix for containers; make prettier

rhtenhove:
This has 2 commits doing 2 separate things in the readme:

1. Fix the container syntax (it has capitals there, and those never work) and move that to option 1 as it works well and the previous option 1 already assumed the availability of Docker - See the first commit for those changes
2. Run [`prettier`](https://prettier.io/) over the markdown, to make it readable without a renderer. Let me know if you like that. I can highly recommend it for anything that it supports, as it makes things nicely readable (except for python, for which they recommend `black`, which does the exact same thing).

### DIFF
--- a/readme_template.md
+++ b/readme_template.md
@@ -68,14 +68,37 @@ folder tree, where you will find the data for `fakenews`, `social`, `gambling`, 
 
 ## Generate your own unified hosts file
 
-You have two options to generate your own hosts file.  You can do it in your own environment, or within a Docker container.  We'll cover Docker first because it's a short section.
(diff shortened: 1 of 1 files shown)


StevenBlack:
Thank you @rhtenhove!

Merging.

rhtenhove:
Cool!

I’m not sure how you generate your readme, it seems that hasn’t happened yet

StevenBlack:
@rhtenhove the readme.md files are updated when I regenerate the hosts files, typically with `[makeHosts.py](https://github.com/StevenBlack/hosts/blob/master/makeHosts.py)`.

Issue #2011, opened 2022-07-15
domain and subdomains cisecurity.org blocked


valerifabio:
Hi, consider to remove from blocklists *.cisecurity.org (internet security).
Regards

StevenBlack:
Closing.  See the discussion in #2102.

PR #2012, opened 2022-07-15
Removed all cisecurity.org subdomains from blocklists

sonarypt:
Cleaned all *.cisecurity.org subdomains from issue #2011.


StevenBlack:
Thank you for this @sonarypt.

All the files you changed in this PR are generative.  This is all explained in the readme which, had you read and understood, would have saved all this wasted effort.

More to the point though, why do you think `cissecurity.org` domains are listed here?  Since you appear to be affiliated with `cissecurity.org`, according to your Github profile, I'm sure you can tell me.

Go ahead.  I'll wait right here.

Closing this PR.

PR #2022, opened 2022-07-25
Fix breakage for Linux

### LINKED ISSUE #2015: hosts file not copying/replacing file in /etc/hosts on Archlinux
Hi, as stated when i do the following command in archlinux the hosts file does not get replaced with the generated one? Ive used the latest release of steven black hosts and press Y to replace hosts in /etc/hosts but nothing happens. It does generate a hosts file but it stays in the folder in PWD.

`python3 updateHostsFile.py`

DPTJKKVH:
Partially revert commit https://github.com/StevenBlack/hosts/commit/0a99194b2844bbc026ab62da69798fdc34815a79 to fix #2015

### DIFF
--- a/updateHostsFile.py
+++ b/updateHostsFile.py
@@ -1312,7 +1312,7 @@ def move_hosts_file_into_place(final_file):
             f"Replacing {target_file} requires root privileges. You might need to enter your password."
         )
         try:
-            subprocess.run(SUDO + [f"'cp {filename} {target_file}'"], check=True, shell=True)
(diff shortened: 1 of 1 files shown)


DPTJKKVH:
Failed checks are because of `./updateReadme.py:65:121: E501 line too long (130 > 120 characters)` which is not part of my change.

StevenBlack:
Thank you for chasing this down @DPTJKKVH.  Awesome!  Merging!

This fixes #2015.

PR #2027, opened 2022-07-27
Add MacOS to check to sudo overwrite hostfile

gprestes:
Someone removed the check for MacOS making the replace step be skipped if the OS was Darwin (MacOS). This PR adds a check if the OS is Darwin and then asks for sudo pwd so the hostfile is replaced by the generated one.


StevenBlack:
Thank you Guilherme @gprestes.  Good catch!

Merging.

PR #2040, opened 2022-08-13
telemetry.acronis.com

FadeMind:
I found it in logs. 
If you using acronis software, then app in background collect telemetry data. IMHO can be blocked and improve privacy. 
Ping every hour. 
https://domain.glass/telemetry.acronis.com


StevenBlack:
Thank you Tomasz @FadeMind.  Merging!

PR #2046, opened 2022-08-18
Adds eBay Phishing domains

dwin:
Received message:

> Hey, I can't afford it, and despite my best efforts over the past ten minutes, there is a problem. Since this is my first purchase in months, I contacted customer service, who informed me that although my account was fine, the issue was that your account doesn't accept purchases from users who haven't made any recently. They provided me with a link and instructed me to send it to you so you could activate all payments and permit both me and other users to make purchases.
> 
> "copy and paste in your browser"
> 
> http://my-payments-disable.store/
> 
> -Kylie

which ultimately redirect to a phishing site: ebay.com.support-myhub.store 

Screen Shot 2022-08-18 at 1 21 39 PM

### DIFF
--- a/data/someonewhocares.org/hosts
+++ b/data/someonewhocares.org/hosts
@@ -1986,6 +1986,7 @@ ff02::3		ip6-allhosts
 0.0.0.0 easyflier.fr
 0.0.0.0 easytic.fr
 0.0.0.0 eauchan.fr
+0.0.0.0 ebay.com.support-myhub.store # eBay phishing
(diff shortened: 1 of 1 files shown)


StevenBlack:
Thank you for this Darwin @dwin.

The file you modified is `data/someonewhocares.org/hosts` which gets clobbered each time we fetch and store this file during a release.  See file `update.json` in the same folder.

The file you can change is `data/stevenblack/hosts` — this is my own ad-hoc list, one that we control.

Closing.

PR #2076, opened 2022-09-05
fake tv stream sites

FadeMind:
Some polish fake tv stream sites. Some links on sites are risky. Other want payment via SMS premium fot grant access to unknown content quality,


StevenBlack:
Thank you Tomasz @FadeMind. Merging.

PR #2123, opened 2022-10-29
docs: Use reproducible Nix expression

l0b0:


readme.md
```diff
@@ -392,20 +392,25 @@ in [::pf4public](https://github.com/PF4Public/gentoo-overlay) Gentoo overlay
...
 
 ```nix
 {
-  networking.extraHosts = let
-    hostsPath = https://raw.githubusercontent.com/StevenBlack/hosts/master/hosts;
-    hostsFile = builtins.fetchurl hostsPath;
-  in builtins.readFile "${hostsFile}";
+  networking.extraHosts = builtins.readFile (
+    builtins.fetchurl {
+      url = "https://github.com/StevenBlack/hosts/raw/3.11.26/hosts";
```
funilrys, line 401: What is the benefit of this change?

StevenBlack:
Hi Victor @l0b0 thank you for this pull request.

If you could answer the question from Nissar @funilrys that would be great.

Take note: we have 16 different main `readme.md` files in this repo, one for each variant.  All those `readme.md` are generated from the `readme_template.md` file in the root of this repo.  So that's where this change should go, presuming it's justifiable on merits. 👍🏻

readme.md
l0b0: Pros:

- _Reproducibility_ is the main one. Anyone using this will get the same result, making debugging much easier than the original code.
- Stability - if the current `hosts` file is working well for you, why take the risk of updating it every time you rebuild the system?
- Safety - if someone ever manages to introduce some malicious content into the `hosts` file it's likely to be detected and fixed by the next release.
- Simplicity - the nested `builtins.readFile (builtins.fetchurl {params}{);` syntax I would expect to be more familiar to most users than the `let … in …` syntax.

Cons:

- Staleness - an old release of the `hosts` file is probably less efficient than a new one at blocking things the user is likely to encounter.
- Maintenance - as a consequence of the above, `configuration.nix` should probably be updated once in a while.

@StevenBlack I'll port this to `readme_template.md` if you want to include this change. I guess it'll be easy to amend that to bump the `url` and `sha256` to the latest release when generating the other files.

StevenBlack: Victor @l0b0 that's fine, but I'm not onboard with hard-coding versions in our `readme.md` files.

StevenBlack:
Seeing no followup, I am closing this.

PR #2163, opened 2022-12-04
Add malware domains

Bertware:
These domains have been discovered in a malware powershell script. 

The first set of domains contains TXT records with powershell code. Some of them are dormant/backup, at least one of them contains an active payload. The powershell payload in turn points to another domain which distributes another payload to specific targeted machines.


StevenBlack:
Thank you for this @Bertware. 

This is a perfect pull request 🌟 motivation and methods are clear, with 51 new domains with zero overlapping with our existing amalgamated list.

Merging!

PR #2169, opened 2022-12-12
Clean-up batch scripts

EmilySeville7cfg:
- use `call :{{label}}` instead of `goto {{label}}`
- terminate all procedures with `exit /b 0`
- use variables where appropriate
- reduce spaghetti code
- existing variable names and label names were not changed

This is a first PR aiming to clean-up script related stuff.

⚠️ Please check whether scripts work yourself, edits were based on my Batch script knowledge but were not tested as I have no Windows machine.

PR #2170, opened 2022-12-12
refresh: `stats.sh` script

EmilySeville7cfg:
- fix first command `echo`: wrong \n usage
- use `read` to put items in array instead of relying on word splitting
- quote variables where suggested by shellcheck


StevenBlack:
Closing.  See #2169.

PR #2196, opened 2023-01-11
Fix GitHub ci workflow badge

EFord36:
See the linked GitHub issue from the previous version of the badge: https://github.com/badges/shields/issues/8671

In summary, these badges were changed in a breaking fashion such that the badges were just linking to the github issue rather than showing the data. Updating the url resolves this.

Also update the link of the badge to specify the workflow file rather than the name, which matches the new behaviour of the badge.

Let me know if I need to 'build' the generated README files or anything.


StevenBlack:
Thank you Elliot @EFord36. This is very cool and thank you for fixing this.

Merging.

PR #2248, opened 2023-03-03
Fix yinzcam apps not loading

alex-taffe:
I work for YinzCam. Any config, cards, or resource hosts are entirely required for app functionality.
For example, no icons will load without resources: https://resources-us.yinzcam.com/nhl/shared/logos/nhl_stl_dark.png

The cards server dictates page layout, the config server handles things like login/ticketing/etc. Those two will entirely prevent the Seahawks app from working. The resources one is preventing assets from loading in over 150 apps


StevenBlack:
Hello Alex @alex-taffe thank you for this PR.

As explained in-detail in the readme, the hosts files here are amalgamated and regenerated at each release.   We're mostly an aggregator of hosts files from active and reputable curators of hosts files, and we package these amalgamated hosts files in various ways.

So the file you modified in this PR isn't going to change anything because that particular file is volatile 😄 

These `yinzcam.com` domains come to us via the AdAway block list. AdAway is in a suspended-animation state right now. Its owners are looking for maintainers. Therefore we've stopped actively updating from AdAway's source and we're operating with our cached copy of their hosts file.  As time goes on, our cached copy gets further out of date.

This might be TMI but we do this in a data-driven way using AdAway's `update.json` configuration file located [here in this repo](https://github.com/StevenBlack/hosts/blob/master/data/adaway.org/update.json).



### THEREFORE...

If you re-submit a new PR that modifies the file [/data/adaway.org/hosts](https://github.com/StevenBlack/hosts/blob/master/data/adaway.org/hosts) then I would entertain merging that.

Closing.

alex-taffe:
My bad, will create a new one. Was just trying to debug this issue middle of the work day.

PR #2249, opened 2023-03-03
Fix yinzcam apps not loading

alex-taffe:
I work for YinzCam. Any config, cards, or resource hosts are entirely required for app functionality.
For example, no icons will load without resources: https://resources-us.yinzcam.com/nhl/shared/logos/nhl_stl_dark.png

The cards server dictates page layout, the config server handles things like login/ticketing/etc. Those two will entirely prevent the Seahawks app from working. The resources one is preventing assets from loading in over 150 apps.

Created to fix #2248


StevenBlack:
Thank you Alex @alex-taffe. This is a good PR.

Merging.

PR #2254, opened 2023-03-10
Remove hits.sh

### LINKED ISSUE #2252: Why block hits.sh ?
Hello. I am the author of the [https://hits.sh/](https://hits.sh/).
I noticed that #2095 this issue makes my site to be blocked.

The guy who created that issue(#2095) said:
> In reality this just means IP harvesting and logging. It's a site that gets loaded silently in the background, unbeknownst to users visiting pages. I'd like to propose we add it to the host list.

Sadly, it's not true. I don't harvesting IP address and logging.
I just simply do `count = count + 1` and return the badge image inside of my logic. [This](https://github.com/silentsoft/hits/blob/dev/src/main/java/org/silentsoft/hits/service/HitsService.java) is the actual source code that I've described.

I'm pretty sure that all of my users who uses "hits.sh" wants to count and **display** the badge image to website to visitors as **their** content.

Please remove "hits.sh" from hosts.

Thanks,

silentsoft:
Fix #2252

### DIFF
--- a/data/StevenBlack/hosts
+++ b/data/StevenBlack/hosts
@@ -2300,9 +2300,6 @@
 # Added September 12, 2022
 0.0.0.0 eu-commercialb.com
 
-# Added September 23, 2022
-0.0.0.0 hits.sh
-
 # Added September 29, 2022
 0.0.0.0 gratlfyingdates.net
 0.0.0.0 www.gratlfyingdates.net


StevenBlack:
Thank you for this pull request Hyesung @silentsoft.

Merging.

This will be live in the next release.

PR #2275, opened 2023-04-04
Adds Echap stalkerware IOC list

Te-k:
Hi,

Here is a PR to add [Echap stalkerware indicators](https://github.com/AssoEchap/stalkerware-indicators) related to #2063. If I am right the other files in the source folders are generated automatically so I only added the `update.json` and hosts files.

Let me know if anything is not in the correct format.


StevenBlack:
Hello @Te-k thank you for this.

I'm doing a quality check, and I'm tracing the inclusion of `account.childsafetytrackerapp.com` in the list, which I consider a Type I errror because it's a client login UI endpoint.

It seems to come from `[stalkerware-indicators](https://github.com/AssoEchap/stalkerware-indicators/tree/1b8e11df9f6933ec025b89e5fc44c3febc26ac98)/ioc.yaml` which, in turn, seems to come from EasyList?  There's also mention of `childsafetytrackerapp.com` in a rules file.

Briefly, how does this work internally?

Te-k:
Basically we are tracking and analyzing stalkerware apps, and maintaining this [ioc.yaml](https://github.com/AssoEchap/stalkerware-indicators/blob/master/ioc.yaml) as the main database for indicators. You can see that there are different types of indicators : packages, certificates, and for network we differentiate websites which are used by the stalkerware to advertise and c2 domains / IPs used for the app to communicate and send data. We have information on both stalkerware and what we call watchware, the key element there is if the app is allowing to hide its presence on the device (if they do, they are listed as watchware and not added to our blocklists)

Then we have [a script](https://github.com/AssoEchap/stalkerware-indicators/blob/research/scripts/generate.py) that generates the different output formats, including host files.

In that case `account.childsafetytrackerapp.com` is a domain used by the app EasyLogger according to our data. I can double check if it is a mistake or not, it is possible that the app is communicating with the UI endpoint (it's pretty common for these kind of apps)

Te-k:
Too bad

PR #2279, opened 2023-04-08
Add more host names to block ads from akamai

AdityaGarg8:


hagezi:
Just a hint:

I would be very careful about adding Akamai deep CNAMES. With the possibility of multiple use, they can restrict functions in completely different places.

This is also the reason why a*.ds*.akamai.net is not on any common blocklist.

PR #2280, opened 2023-04-08
CI: limit pushes to master

XhmikosR:


StevenBlack:
Thank you as always @XhmikosR this is good!  Merging.

PR #2281, opened 2023-04-08
Update .editorconfig

XhmikosR:


.editorconfig
```diff
@@ -1,14 +1,20 @@
...
-[*.py]
+[*]
+charset = utf-8
+end_of_line = lf
+indent_size = 2
 indent_style = space
+insert_final_newline = true
 trim_trailing_whitespace = true
-indent_size = 4
-end_of_line = LF
```
StevenBlack, line 7: @XhmikosR as far as you know, will this change on L7 affect Windows users?

StevenBlack: @XhmikosR Everything else LGTM.

XhmikosR: It will still be LF inherited from the first block like it was before

StevenBlack: @XhmikosR Duh! Of course. Thanks.

StevenBlack:
Thank you @XhmikosR!

Merging!

PR #2282, opened 2023-04-08
Update npm deps and regenerate lockfile

XhmikosR:
Closes #2278 while at it.


StevenBlack:
@XhmikosR yup!  This is good!

Merging.

PR #2325, opened 2023-05-14
PlayStation Telemetry over API

DavidGarciaCat:
Just found these two entries when I checked my Pihole after playing on my PS


DavidGarciaCat:
There's another entry ( `takedown.api.playstation.com` ) that I don't know if it should be also included.

DavidGarciaCat:
There's another entry (`takedown.api.playstation.com`); I don't know if it should also be included.

StevenBlack:
Thank you for this PR David @DavidGarciaCat!

About the third domain, if we aren't sure, then let's not.

Merging!

DavidGarciaCat:
hey @StevenBlack 

I just realised the two new items were only added to the main `hosts` file, not the alternatives. I guess it makes sense, and then they are merged somehow, or shall I create a new update for them too?

Thanks, and sorry for the silly question

StevenBlack:
David @DavidGarciaCat the new domains are merged into the repo, and are part of the data, but won't propagate until the next release when I'll build all our lists.

PR #2330, opened 2023-05-19
Add more host names to block ads/popups from netu.tv videoplayer

BaxAndrei:

PR #2332, opened 2023-05-20
Add more host names to block ads/popups from netu.tv videoplayer

BaxAndrei:
https://github.com/StevenBlack/hosts/pull/2330


StevenBlack:
Thanks Andrei @BaxAndrei!

Merging.

PR #2352, opened 2023-06-08
Attempted to use the new GitHub Linguist hosts syntaxing

DandelionSprout:
If I understand .gitattributes values correctly, this should hopefully result in colour text on e.g. https://github.com/StevenBlack/hosts/blob/master/data/mvps.org/hosts

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -5,3 +5,5 @@
 
 *.py linguist-detectable=true
 *.nb linguist-detectable=false
+
+hosts linguist-detectable linguist-language=hosts


funilrys (approved):

StevenBlack:
Hello @DandelionSprout thank you!  Let's try this!

Merging.

StevenBlack:
@DandelionSprout I'm not seeing any benefit to this.  There's no change in how Github displays hosts files, and in VSCode the changes I see are of no benefit whatsoever. 

Moreover the extension evidently doesn't recognize `0.0.0.0` as an IP address?

I'm reverting this.

However, thank you for the suggestion.

Issue #2353, opened 2023-06-09
Adding Google Colab option to README


ScriptTiger:
Thoughts on adding a Google Colab option to the README for generating your own hosts file? I think it might even beat Docker as far as simplicity and ease of access.

https://colab.research.google.com/drive/1tYWXpU2iuPDqN_o03JW9ml3ExO80eBLq?usp=sharing

StevenBlack:
Hi @ScriptTiger. Thanks! Pull requests speak loudly 😄 

I'll probably be merging PR #2335 soon so please watch for merge conflicts.

ScriptTiger:
Roger! Just thought I'd send out some feelers first. I'll send a PR later. I'll keep the footprint as minimal as possible to reduce conflicts, basically just a note in the "Option 3: Generate it in your own environment" section. I'm not going to presume it to be the best option off the bat, so I'll just leave it there and if others want to rearrange the order of preference, that can be a discussion later.

PR #2354, opened 2023-06-11
Add note and link for Google Colab to README

ScriptTiger:
Follow-up to issue #2353.

### DIFF
--- a/readme_template.md
+++ b/readme_template.md
@@ -138,6 +138,8 @@ docker run --rm -it stevenblack-hosts updateHostsFile.py
 
 ### Option 3: Generate it in your own environment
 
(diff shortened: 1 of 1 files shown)


StevenBlack:
Hi @ScriptTiger thank you for this.

I'm not happy about this being the first thing that follows `### Option 3: Generate it in your own environment`.  This is not related to Option 3, it's a whole different option, a different environment, one that isn't "in your own environment".

**THEREFORE** can you please present this as **Option 4**

**ALSO** the text you use makes no sense.  Here's what's written in this PR:

> You can also alternatively use either your own local or remote environment, such as quickly and easily spinning up a free Google Colab.

That's babblelspeak. It makes no logical or lexical sense.  For example, how many unnecessary babblewords are in this sentence segment: "***You can also alternatively use either...***"?

Please clean this up. Thanks!

ScriptTiger:
Option 4 added.

StevenBlack:
Thank you @ScriptTiger. This is pretty cool.

Merging!

PR #2364, opened 2023-06-14
Update hosts

hkoster90:
added fairspin.io


StevenBlack:
Thank you for this @hkoster90 but I must decline this pull request.

I believe this is well documented in the [repository readme](https://github.com/StevenBlack/hosts/blob/master/readme.md): we generate hosts files by combining lists from reputable sources, and we package these products in various ways.

The file you submitted here is generative. It gets clobbered and re-created with each release.

I just added this new diagram to the readme — just a few moments ago — and hopefully this clarifies it.

aggregator

Closing.

PR #2416, opened 2023-08-19
Added more porn sites

jdiezvenegas:
Hi,
I made a porn list some months ago. I used it to block porn sites in qutebrowser. Maybe there are some new sites. At the end of the list I added yours. There´s also a source list.


StevenBlack:
Thank you for this @jdiezvenegas.

The changed files in this PR are generative which means, they get regenerated from data at each release.  So these changes are not going to stick.

Secondly, I'm not sure what you've done here, so I respectfully decline this PR.  

The role of this list isn't to block all the pron in the world, but merely to keep porn out of one's house, or one's company/organization.  Which we do just fine right now.

PR #2421, opened 2023-08-24
flake: rm flake-utils dependency

toastal:
Pulling in an entire dependency to call a for-loop is wasteful & largely useless.

When user adds this module to their config, flake-utils & all of its subdependencies will be pulled into the user’s flake.lock file. This for-loop was only being used for the developer shell to which a lot of folks probably aren’t doing active developments in this project as the module itself doesn’t require it. Potentially damagingly is that this project lacks its own flake.lock so the latest flake-utils will always be downloaded regardless of if it that version is compatible or not. Additionally, flake-utils’ default system list doesn’t include i686-linux which upstream Python3 in Nixpkgs does.

The alternative solution to these problems is to remove the dependency & just write a for-loop in this project. This solution could be more or less robust, but it is an extensible version of that loop that could handle overlays or config changes if needed in the future.

(the diff looks bigger than it really is because of the module block needed to be indented for the new `let` block; a reviewer can look at the [whitespace-free diff](https://github.com/StevenBlack/hosts/pull/2421/files?diff=unified&w=1))

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -1,45 +1,59 @@
 {
   description = "Unified hosts file with base extensions.";
-  inputs.flake-utils.url = "github:numtide/flake-utils";
-  outputs = { self, nixpkgs, flake-utils }: {
-    nixosModule = { config, ... }:
-      with nixpkgs.lib;
-      let
-        cfg = config.networking.stevenBlackHosts;
-        alternatesList = (if cfg.blockFakenews then [ "fakenews" ] else []) ++
-                         (if cfg.blockGambling then [ "gambling" ] else []) ++
-                         (if cfg.blockPorn then [ "porn" ] else []) ++
-                         (if cfg.blockSocial then [ "social" ] else []);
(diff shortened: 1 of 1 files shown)


StevenBlack:
Hi @toastal thank you for this.

Here's why we use Flake in this project: https://github.com/StevenBlack/hosts/pull/1776

Closing.

toastal:
@StevenBlack This isn’t removing the flake. I use the Flake in my config.

This merge request is to remove an unneeded dependency that all downstream users must download. The dependency can be inlined with a couple of lines of Nix.

toastal:
@Deleh , as the author of the merge request link, Could I get confirmation that this doesn’t break anything?

Deleh:
> @Deleh , as the author of the merge request link, could I get confirmation that this doesn’t break anything? If it doesn’t I will reopen.

The changes don't break anything, the module and dev shell still work as expected without the dependency :+1:

PR #2431, opened 2023-09-05
Fix analytics.shein.* lines not having an IP at all

nabijaczleweli:
two invalid lines


StevenBlack:
Dziękuję @nabijaczleweli!

Merging.

PR #2433, opened 2023-09-07
Correct 2nd Iter: Hostnames to lowercase.

funilrys:
As mentioned by @StevenBlack in #2400, hostnames should be converted to lowercase.


StevenBlack:
Thank you as always Nissar @funilrys! 🌟

Merging!

PR #2434, opened 2023-09-08
Correct 2nd Iter: Ignore subject with slashes (/).

funilrys:
Upon checking the latest release after merging #2433, I noticed that some subjects with slashes go/went through.

This patch fixes that by ignoring subjects that contain slashes.


StevenBlack:
Thank you Nissar @funilrys.  Good catch.

Merging.

PR #2444, opened 2023-09-13
Added information about Hosts Compress Windows to the README; this to…

Lateralus138:
…ol is recommended over the previous scripts. - 09/13/2023 20:02:32 UTC

I considered removing the Hosts Compression Scripts link, but I think it should stay just in case someone prefers the (much slower) script for whatever reason. I will be adding a Linux C++ version soon. I've decided to separate the Windows and Linux versions unlike some of my other tools.

Thanks for any consideration of this addition.

### DIFF
--- a/readme_template.md
+++ b/readme_template.md
@@ -453,7 +453,7 @@ the
 [the comments within the `cmd` file](https://github.com/StevenBlack/hosts/blob/master/disable-dnscache-service-win.bat)
 for more details.
 
-Disabling the DNS Cache Service can cause issues with services and applications like *WSL* and it's possible to compress the hosts file and negate the need to disable the DNS caching service. You can try the PowerShell compression script and check out the guide located at the [Hosts Compression Scripts repository](https://github.com/Lateralus138/hosts-compression-scripts).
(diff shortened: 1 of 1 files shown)


StevenBlack:
Hello Ian @Lateralus138, thank you for this PR.

Does it matter that this is a "C++ command line tool"?  Why mention this facet at all?

I'm puzzled why we would want... 

* before: "powershell script"
* now: "C++ command line tool"

This seems retrograde to me, upon first blush.  What do you think?

Lateralus138:
> Does it matter that this is a "C++ command line tool"? Why mention this facet at all?

"command line" Because some people prefer GUIs and so people should know what they're working with and "C++" because, I don't know, just a habit I guess. I'd say lots of people are shying away from C++ in preference to Rust as they believe it's "memory safe-r", but I can assure people their are no memory leaks, and of course, this can be checked with Dr. Memory or whatever leak tools people use. 

> I'm puzzled why we would want...
> - before: "powershell script"
> - now: "C++ command line tool

The PowerShell script was thrown together in emergency as countless people are having issues in Windows and a script can be written exponentially faster than an actual program and I wanted to get it out there for people. The actual program was the end goal, but I thought I'd leave the script for people that for whatever reasons would prefer a script. Options and all that.

StevenBlack:
Thank you Ian @Lateralus138.  Merging.

PR #2445, opened 2023-09-14
Create NWHost

Harry753:


StevenBlack:
Thank you for this Harry @Harry753.

Please read the [repository readme](https://github.com/StevenBlack/hosts/blob/master/readme.md) with the goal of better understanding what we do here.

The file you submit in this PR is generative — it is overwritten by the release process.

Please submit this request to [one of our curated sources](https://github.com/StevenBlack/hosts#sources-of-hosts-data-unified-in-this-variant).

Closing.

PR #2461, opened 2023-09-27
Allow api.radar.io requests

pcl:
Radar is a legit business that serves up map tiles.

### DIFF
--- a/data/adaway.org/hosts
+++ b/data/adaway.org/hosts
@@ -8653,9 +8653,6 @@
 127.0.0.1 r-ad.ne.jp
 127.0.0.1 bk.r-ad.ne.jp
 
-# [radar.io]
-127.0.0.1 api.radar.io
-
 # [radiusnetworks.com]
 127.0.0.1 proximitykit.radiusnetworks.com
 
--- a/hosts
+++ b/hosts
@@ -11241,9 +11241,6 @@ ff02::3 ip6-allhosts
 0.0.0.0 r-ad.ne.jp
 0.0.0.0 bk.r-ad.ne.jp
 
-# [radar.io]
-0.0.0.0 api.radar.io
-
 # [radiusnetworks.com]
 0.0.0.0 proximitykit.radiusnetworks.com
 


pcl:
See https://news.ycombinator.com/item?id=37669290

StevenBlack:
Thank you for this Patrick @pcl.

Do you work for Radar?

StevenBlack:
Patrick @pcl I appreciate this pull request but, based on the evidence I see, I decline.

Closing.

pcl:
I don’t.

FYI, according to their privacy FAQ, location tracking is end-user opt-in. However, looking at their privacy policy, the policy claims end-user opt-out. 

https://radar.com/trust
https://radar.com/privacy

Also, unrelatedly — I noticed that some of your hosts files sent the address to 0.0.0.0 and some to 127.0.0.1. I didn’t investigate further to see if you had a reason for the inconsistency.

StevenBlack:
Hi Patrick @pcl all our hosts variants use `0.0.0.0` by default except in a few records at top of file.

That section is verbatim, a unified and de-duplicated collection of entries from OEM hosts files, sourced from a wide variety of operating systems.

Decided not to modify those because hosts files long predate the internet as we know it, and lord knows what scripts exist in userland that may parse hosts files for unimaginable reasons.  So the decision was, let's not gratuitously mess with those.

Note that we provide an [`--ip` flag](https://github.com/StevenBlack/hosts#command-line-options) for people who want to generate something other than our default `0.0.0.0` mapping.

kincaidoneil:
Hello, can this be revisisted? I've had users dissatisfied because this unexpectedly broke part of our web app.

I think there are a couple things worth clarifying:

1. **Radar is opt-in.** Users must explicitly grant geolocation permissions (browser or OS level) or _manually enter their own address_ before Radar can record it. AFAIK, Radar's "tracking" merely is just saving these user-provided coordinates and device IP to their backend, which is pretty similar to the website/app just using a different mapping provider and saving them to its own backend.
2. Radar is for **first-party** location use cases, think like a DoorDash-style delivery app that needs realtime location, or a dating app that asks users for their location to show nearby people. Importantly, again, these are things the user already must opt into at the platform level

PR #2475, opened 2023-10-03
Add porn entries

scorptiodnelz:


StevenBlack:
Thank you for this @scorptiodnelz.

This PR adds domains to a data source that isn't limited to porn. Our base list does not block porn, social media, gambling, etc.

This is all explained in the readme 😄 

Closing.

PR #2498, opened 2023-11-07
Turkish National Cyber Incident Response Center UPDATE

alicangnll:
Turkish National Cyber Incident Response Center UPDATE

Turkish National Cyber Incident Response Center, adding (7/24) Turkish web sites.

### DIFF
--- a/data/USOM/readme.md
+++ b/data/USOM/readme.md
@@ -1,0 +1,3 @@
+# Source domain history
+
+This page is USOM (Turkish National Cyber Incident Response Center)'s fake site list
--- a/data/USOM/update.json
+++ b/data/USOM/update.json
@@ -1,0 +1,9 @@
+{
+    "name": "USOM",
+    "description": "Turkish National Cyber Incident Response Center based on [usom.gov.tr](https://www.usom.gov.tr/) content.",
+    "homeurl": "https://www.usom.gov.tr/",
+    "frequency": "occasionally",
+    "issues": "https://www.usom.gov.tr/iletisim",
+    "url": "https://www.usom.gov.tr/url-list.txt",
+    "license": ""
+}


StevenBlack:
Thank you very much for this Ali.  This pull request appears to be technically perfect. You're good! 🤝 

Using [ghosts](https://github.com/StevenBlack/ghosts) here is what I get. I put the hosts file in my local clipboard, then:

```
$ ghosts --clip
----------------------------------------
Base hosts file summary:
----------------------------------------
Location: https://raw.githubusercontent.com/StevenBlack/hosts/master/hosts
Domains: 151,352
Bytes: 4.7 MB
----------------------------------------
Compared hosts from clipboard summary:
----------------------------------------
Location: clipboard
Domains: 243,066
Bytes: 7.6 MB
Intersection: 250 domains
```

This is problematic for many reasons.

1. The goal of this project, as stated in the readme: ***These unified hosts files should serve all devices, regardless of OS.***  Unfortunately, with a combined size almost 400,000 hosts, this will lock-up every Windows machine who uses our hosts file. 
2. Our hosts file, curated by many dedicated people across the planet, is presently 150,000 hosts.  This proposed hosts file consists of 240,000 hosts.  What would be an improbably small overlap of common domains, in your view?  In my view, a 10,000 domain overlap would be improbably small.  But here we have just 250 domains overlapping.  This just doesn't feel right, based on my experience doing this for many years.

Therefore, I must decline this Pull Request.

But I commend you for your effort here. You are awesome. It's just that we can't use this because it's a Windows OS killer, and the list doesn't make much sense on a comparative basis — the overlapping domain count is improbably small, not by a little, but by a lot.

Closing.

Lateralus138:
> Unfortunately, with a combined size almost 400,000 hosts, this will lock-up every Windows machine that uses our hosts file.

I would just liked to point out that if fully compressed `(400000 / 9)` this would reduce it to 44,444 entries which is doable in Windows, but I certainly understand not everyone can, knows how, and/or wants to compress it. Just saying if this was a usable option somewhere some people could deal with it. Just my 2 cents, of course :D

PR #2502, opened 2023-11-13
Added 4 entries to list.

TimTheBig:


StevenBlack:
Thank you for this Tim @TimTheBig.

The file you changed is generated, and gets overwritten by merging hosts in the `/data` folder, which all get refreshed with each release using the Python programs we maintain here.

Therefore the commits in this PR would not survive the release process.

If you want to add domains to our list you can do that in the `/data/StevenBlack/hosts` file.

Fair warning: we've had many discussions about Microsoft and Apple domains over the years, and it's unlikely that this PR will be accepted anyway.  It's complicated, and potential side-effects are far ranging. The chances you've found something that all of us have somehow missed with these domains is pretty close to zero.

Closing.

PR #2526, opened 2023-12-12
Fix return type in prompt_for_update

XhmikosR:
I was just looking at the CodeQL issues on my fork and this one seemed the easier to fix.

I'll leave the rest to someone more proficient with Python :)


StevenBlack:
Thank you @XhmikosR!  It's in.

PR #2529, opened 2023-12-15
Support DNS Flush for Windows in Python

gfyoung:
Title is self-explanatory.  Ran locally to confirm it works.

With this feature now in Python, `updateHostsWindows.bat` becomes unnecessary.


StevenBlack:
Thank you for this @gfyoung.

I feel this modification merits discussions beforehand.

Firstly, we would want to do this because... ?

Secondly, the required [readme_template.md](https://github.com/StevenBlack/hosts/blob/master/readme_template.md) file modifications do not appear to be included in this PR.

Thirdly we're talking about Windows here, with its vast clusterf of version in userland, users who may or may not have admin privileges, how does doing this in Python help at all? In particular, if it fails, what recourse does the end user have? Run the whole `updateHostsFile.py` again? Where are the error messages to help prevent an avalanche of support requests from Windows userland?

At least the `.bat` file is a separate process with fairly clear recourse for the user.  At least, that's my gut feeling. I'm willing to be convinced otherwise.

gfyoung:
> the required [readme_template.md](https://github.com/StevenBlack/hosts/blob/master/readme_template.md) file modifications do not appear to be included in this PR

Good catch.  I'll add those.

> users who may or may not have admin privileges, how does doing this in Python help at all

The Python script has already incorporated admin-level Windows commands (see https://github.com/StevenBlack/hosts/pull/1999).  This PR simply continues what was started.

The script prints out that admin privileges are required for flush DNS cache. (see https://github.com/StevenBlack/hosts/pull/2529/files#diff-0666db8e6c944e3c08fdc8f42f9bf8d85edd3aca630d2fa08d8adb0ebd677f02L1453-L1454)  If the user has no admin privileges, the script will still run.  It just won't flush the DNS cache, which wouldn't be possible anyway because you don't have admin privileges.

The `.bat` file doesn't change that reality and isn't really a recourse.  If it is, I'd raise the question why Windows is the only one that needs this treatment vs. other platforms i.e., replace in your feedback Linux or Mac and clarify why those two don't need their own shell scripts for example.

gfyoung:
@StevenBlack

StevenBlack:
Thank you for those thoughts @gfyoung.

Here's my perspective: well over 90% of support issues here (and in my personal email inbox) are Windows-related.  

Quite frankly, Windows is a pain in the ass. I don't want to do anything that increases support loading from Windows' userland.

This PR strikes me as a solution in search of a problem. There's nothing to gain, but lots to lose.

I appreciate your "works on my machine certification" but there are much wider risks here and presently I'm inclined to decline this PR.

I'll think about it.

gfyoung:
> This PR strikes me as a solution in search of a problem. There's nothing to gain, but lots to lose.

Windows support is subordinate to other operating systems.  The Python script handles everything for non-Windows systems.  I shouldn't have to run both the Python script and `.bat` file for the same level of support.  As it stands, you already merged https://github.com/StevenBlack/hosts/pull/1999, which was arguably more risky than this change.  I'm simply building off of it.

> I appreciate your "works on my machine certification" but there are much wider risks here and presently I'm inclined to decline this PR.

Totally get there could be wider risks, but that's why this is a PR.

Others are welcome to verify themselves that this change works before this gets merged.

XhmikosR:
Actually this change does make sense as long as it still works the same, which I haven't tried since I don't use the hosts this way.

I'd say let's wait until more people on Windows try this and confirm it works the same as the previous solution and then merge it.

PS. I never used an elevated cmd to flush the DNS on Windows. Is it really needed?

gfyoung:
> PS. I never used an elevated cmd to flush the DNS on Windows. Is it really needed?

Good question: I do recall it being required at some point but definitely not needed in more recent versions of Windows.  A good follow-up PR would be to remove the `SUDO` call if testing confirms that it's no longer needed.

winlover32 (approved):

danilogco (approved):

goodevile:
my POV: 

Windows is PAIN.

There should be documentation before implementing this, it's more like a hack than a functionality, why not just add a documentation for Windows users like:

`"if you are a windows user, we have this script that automates this X process, you can read the script and review it , we are not responsible for any unexpected behavior."`

I think that there are more important stuff than creating scripts for automating Windows stuff.

PR #2543, opened 2023-12-28
Added rtnl.bxcl.de

kjde:
TonieBox "Telemetry"

Thanks to great people at [toniebox-reverse-engineering](https://toniebox-reverse-engineering.github.io/docs/wiki/general/protocol-analysis/#rtnlbxclde) it should be a good idea to block rtnl.bxcl.de.

As answered [here](https://media.ccc.de/v/37c3-11993-toniebox_reverse_engineering#t=3141) you can block it without side effect.


StevenBlack:
Thank you for this Kaspar @kjde.

Merging.

PR #2555, opened 2024-01-11
Add android instructions

FahadBinHussain:
Added instructions for how to setup and edit android `hosts` file

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -399,10 +399,15 @@ We tried that. Using `0` doesn't work universally.
 To modify your current `hosts` file, look for it in the following places and
 modify it with a text editor.
 
-- **macOS (until 10.14.x macOS Mojave), iOS, Android, Linux**: `/etc/hosts`
+- **macOS (until 10.14.x macOS Mojave), iOS, Linux**: `/etc/hosts`
   file.
 - **macOS Catalina:** `/private/etc/hosts` file.
 - **Windows**: `%SystemRoot%\system32\drivers\etc\hosts` file.
+- **Android**: `/system/etc/hosts` file.
+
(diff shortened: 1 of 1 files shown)


Hackurei:
you don't need any magisk module to replace this file, you just need root access.
i believe this file is on a read only partition, so you just need to remount the partiton as read/write first.

also, the domain you provided is blocked by ublock origin:
blocked

FahadBinHussain:
> you don't need any magisk module to replace this file, you just need root access. i believe this file is on a read only partition, so you just need to remount the partiton as read/write first.
> 
> also, the domain you provided is blocked by ublock origin: blocked

I don't know if remounting the partition as read/write would work or not. However, when I attempted to change the host file (the device was rooted), the changes were being made but the hosts were not being blocked. Additionally, I tried other methods to achieve this, but none of them were successful except for the Magisk module. Therefore, I thought it would be helpful for others to include that in the readme, as the Android instructions were also insufficient. As for the uBlock Origin domain block, that is merely a false positive.

Lateralus138:
And note that root isn't even necessary to successfully get a hosts file to work on Android, a DNS client like NextDNS works perfectly fine and a lot simpler process. Not to mention the fact that less and less Android devices can be rooted nowadays and the average user doesn't or can't root their device.
Here's screenshots of 3 different browsers, one with 2 different links from Steven's main hosts file using NextDNS on Android:
android_with_hosts

StevenBlack:
Thank you for this PR Fahad @FahadBinHussain, I appreciate the effort here.

Given the comments above ⬆️ I decline to include these changes.

Closing.

PR #2623, opened 2024-04-03
Minor whitespace/line endings consistency changes

XhmikosR:
Non-whitespace diff: https://github.com/StevenBlack/hosts/pull/2623/files?w=1


StevenBlack:
Thank you for this @XhmikosR.

A question: how does this square with our [`.editorconfig`](https://github.com/StevenBlack/hosts/blob/master/.editorconfig)?

Also, there's way more than whitespace and consistency changes in this PR. What's the motivation for those changes? Any downsides?

XhmikosR:
I actually applied these in compliance with the current .editorconfig settings. It seems that previously these files landed with inconsistent settings.

XhmikosR:
> Also, there's way more than whitespace and consistency changes in this PR. What's the motivation for those changes? Any downsides?

Just moving things around nothing fancy. I can revert those changes if you prefer :)

.github/dependabot.yml
```diff
@@ -1,17 +1,17 @@
 version: 2
 updates:
-  - package-ecosystem: "pip"
+  - package-ecosystem: "github-actions"
```
funilrys, line 3: Is there a reason behind the order change ?

.github/workflows/depsreview.yaml
```diff
@@ -13,5 +13,5 @@ jobs:
         with:
           persist-credentials: false
 
-      - name: 'Dependency Review'
+      - name: Dependency Review
```
funilrys, line 16: If we start like this, then we should remove all quotes that are unnecessary 😉

XhmikosR: Well, it's without quotes already in most cases hence the change

.github/dependabot.yml
XhmikosR: https://github.com/StevenBlack/hosts/pull/2623#issuecomment-2034699665

I just keep actions first and then any package managers

.github/workflows/depsreview.yaml
funilrys: I mean in all other yaml files 😉

StevenBlack:
Hi @XhmikosR and @funilrys!

I'm happy 😄 and I'm merging this now.

The whole xz thing has me on-edge.  Not that @XhmikosR would be a vector 🤣 — I've decided to be much more sus generally, starting now, and in the future.

PR #2625, opened 2024-04-08
notepad.plus

FadeMind:
Ref https://notepad-plus-plus.org/news/help-to-take-down-parasite-site/

### DIFF
--- a/data/StevenBlack/hosts
+++ b/data/StevenBlack/hosts
@@ -2734,3 +2734,6 @@
 
 # Added April 3, 2024
 0.0.0.0 canadapost.ca-help-h.top
+
+# Added April 8, 2024
+0.0.0.0 notepad.plus


StevenBlack:
Thanks Tomasz!

Merging!

iam-py-test:
[The website clearing discloses the fact that it is not owned by NotePad++'s developer(s)](https://www.bleepingcomputer.com/news/security/notepad-plus-plus-wants-your-help-in-parasite-website-shutdown/), albeit at the bottom of the page where most people never look:
> Notepad++ is a trademark of Don Ho. Notepad.plus is not affiliated with, sponsored by, or endorsed by Don Ho in any way. This is an unofficial fan website created for general information/educational purposes only. Any information found in this website is our personal opinions and do not purport to reflect the opinions or views of Don Ho or his representatives. All other trademarks are the property of their respective owners.

The download link goes to the legitimate website, and while the website does have some unsavory trackers (Google, Yandex), the closest thing to an ad is a request to `https://mc.yandex.ru/metrika/advert.gif` - and even that's just a tracking pixel. There are no visible advertisements that I see (tested in both Firefox and Chrome with uBo and tracking protection disabled).
I see no evidence that it is
> riddled with malicious advertisements on every page

All of that said, my security software blocklists notepad[.]plus, so maybe they know something I don't.
Thanks

FadeMind:
@iam-py-test 
@donho is not owner of notepad[.]plus
The download link may be swapped with an unsafe one containing malicious code. A few months of gaining trust - oh, after all, it's just a link to Notepad++, and then one line of code is swapped and multiple computers are infected. Who's to blame? There is a note that DonHo is not the owner, but who has time to read the fine print? The site is misleading. Simply put.  
In short: Notepad[.]plus domain is a scam and potentially phishing/malware site. No exceptions.

donho:
More info here:
https://community.notepad-plus-plus.org/topic/24001/when-google-is-possessed-by-profit/12?_=1712912546094

iam-py-test:
With all due respect, currently safe websites should not be blocked based on hypothetical scenarios. There is currently no evidence the website is harmful to users, nor is there any evidence of deceptive advertising (or for that matter any advertising) anywhere on the website. If you have evidence to the contrary, or if the website changes, then it should be blocklisted.
> donho is not owner of notepad[.]plus

That does not make it malicious. Nowhere does the website claim to be owned by the actual maintainer of NotePad++, and while the disclaimer should be much more prominent*, it does exist. Moreover, at the risk of using whataboutism, [it is no less prominent than other fan websites](https://ublockorigin.com/).

Thank you

\* I have contacted the website owner and requested they make it more visible. While I doubt I will receive a response, I believe the website owner should nevertheless be given a chance to rectify this issue.

FadeMind:
@iam-py-test 

> With all due respect, currently safe websites should not be blocked based on hypothetical scenarios.

Owner site think same. 
Popular Name Mark, Popular Text Editor Tool, "legit" domain name.
Possibly RISK and danger is high. Google now blocked it. This is good. No more visible in search results. This is good too. 
Download aggregators  like softsonic using popular names in subdomain, but at end they are adware sites with risky self hosted downloader. 
Anyone can write a  blog, BAD is using similar name. 
Phishing site work like this. Period.

PR #2661, opened 2024-05-30
Add netflix log host

JMaravalhasSilva:
This PR adds the logs.netflix.com host, which appears to be some sort of tracker.

I've tested the netflix app on webOS and on a normal browser and it works fine with this host blocked.


StevenBlack:
Thank you for this PR José @JMaravalhasSilva.

Merging.

PR #2690, opened 2024-07-18
add pubfuture domains

wbedu:
Adds Pubfuture.com ad network and its subdomains.


StevenBlack:
Thank you for this Will @wbedu.

In our base list we block domains used for adware, malware, or worse.

This PR includes several records that are related to management and apis, which are not malicious.

Based on this I decline this PR.  But nonetheless thank you for submitting this.

Closing.

PR #2713, opened 2024-08-31
removed trailing dot

jworrell:
grid-bidder.criteo. has a trailing period, while no other domain does


StevenBlack:
Justin @jworrell! Good catch!  Thank you!

Merging.

PR #2731, opened 2024-09-23
remove cdn.cookielaw.com

C0nsultant:
This breaks multiple sites, most importantly certain subpages of SoundCloud when logged in (notifications, settings, ...).

This is the loader CDN for a cookie-banner.
It is in my opinion none of the following:
- adware
- malware
- fakenews
- gambling
- porn
- social

While the cookie banners certainly can be classified as `annoyances`, this list does not seem to be about annoyances.

### DIFF
--- a/data/StevenBlack/hosts
+++ b/data/StevenBlack/hosts
@@ -2495,7 +2495,6 @@
 
 # Added March 7, 2023
 0.0.0.0 a125375509.cdn.optimizely.com
-0.0.0.0 cdn.cookielaw.org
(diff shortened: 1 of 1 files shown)


StevenBlack:
Hello @C0nsultant thank you for this.

[We've talked about cookielaw many times before](https://github.com/StevenBlack/hosts/issues?q=is%3Aissue+cookielaw+is%3Aclosed), and this isn't going to change.

If SoundCloud can't function when cookielaw is unreachable, that's on SoundCloud; it's their bug.

Closing.

Issue #2719, opened 2024-09-05
[REQ FEATURES] Alternative raw without 0.0.0.0


alsyundawy:
**[REQ FEATURES] Alternative raw without 0.0.0.0**

like example

**https://github.com/StevenBlack/hosts/tree/master/alternates/fakenews-gambling-only**
`100percentfedup.com
20minutenews.com
21stcenturywire.com
365usanews.com
4threvolutionarywar.wordpress.com
5galert.com
70news.wordpress.com
aanirfan.blogspot.com
abcnews.com.co
abilenetimes.com
abolishsocialism.com
abortions.news
abqtimes.com
absurd.news
actualidad.rt.com
adairmadisonnews.com
adamscountytimes.com
adareporter.com
addictinginfo.org
addiction.news
adobochronicles.com
africannewsupdates.com
ahtribune.com
aikentimes.com
akbusinessdaily.com
akronreporter.com
albanystandard.com
albusinessdaily.com
alertchild.com
aljazeera-channel.com
alleghenyhighlandstoday.com
allnewspipeline.com
almastandard.com
alohastatenews.com
altleft.news
altoonatimes.com
aluminum.news
alynews.com
`

StevenBlack:
This is a good suggestion @alsyundawy.

Can we do this upon generation with the `--ip` flag?  I'll check.  if not, this is certainly an easy get.

StevenBlack:
Offhand I think we'll code this extending the `--ip` flag to accept `--ip none`.

alsyundawy:
yes is call *-domain-only

ScriptTiger:
I just wanted to tag a related issue, which has apparently been open for years. As soon as this is resolved, we can knock that one out, as well. Or we could just proactively knock that one out now as being stale.
https://github.com/StevenBlack/hosts/issues/1111

PR #2756, opened 2024-10-16
support alternative raw without 0.0.0.0 

ldlac:
Support this request #2719 

how it should be used `python updateHostsFile.py -a -g -i None`

I made it that it cannot be combined with -c or -m, should it be?

### DIFF
--- a/updateHostsFile.py
+++ b/updateHostsFile.py
@@ -51,6 +51,8 @@ else:
 
 # Project Settings
 BASEDIR_PATH = os.path.dirname(os.path.realpath(__file__))
+HOSTS_FILENAME = "hosts"
+DOMAINS_FILENAME = "domains.txt"
 
 
 def get_defaults():
@@ -262,6 +264,15 @@ def main():
     settings["targetip"] = (
         None if str(settings["targetip"]).lower() == "none" else settings["targetip"]
     )
+    settings["skipstatichosts"] = (
+        settings["targetip"] is None or settings["skipstatichosts"]
+    )
+    settings["keepdomaincomments"] = (
+        settings["targetip"] is not None and settings["keepdomaincomments"]
+    )
+    options["outputfilename"] = (
+        HOSTS_FILENAME if settings["targetip"] is not None else DOMAINS_FILENAME
+    )
 
     update_sources = prompt_for_update(freshen=settings["freshen"], update_auto=auto)
     if update_sources:
@@ -291,19 +302,22 @@ def main():
     merge_file = create_initial_file(
         nounifiedhosts=no_unified_hosts,
     )
-    remove_old_hosts_file(settings["outputpath"], "hosts", settings["backup"])
+    remove_old_hosts_file(
+        settings["outputpath"], options["outputfilename"], settings["backup"]
+    )
+
+    final_file = open(
+        path_join_robust(settings["outputpath"], options["outputfilename"]), "w+b"
+    )
+    temp_file = tempfile.NamedTemporaryFile()
+    remove_dups_and_excl(merge_file, exclusion_regexes, temp_file)
+
     if settings["compress"]:
-        final_file = open(path_join_robust(settings["outputpath"], "hosts"), "w+b")
-        compressed_file = tempfile.NamedTemporaryFile()
-        remove_dups_and_excl(merge_file, exclusion_regexes, compressed_file)
-        compress_file(compressed_file, settings["targetip"], final_file)
+        compress_file(temp_file, settings["targetip"], final_file)
     elif settings["minimise"]:
-        final_file = open(path_join_robust(settings["outputpath"], "hosts"), "w+b")
-        minimised_file = tempfile.NamedTemporaryFile()
-        remove_dups_and_excl(merge_file, exclusion_regexes, minimised_file)
-        minimise_file(minimised_file, settings["targetip"], final_file)
+        minimise_file(temp_file, settings["targetip"], final_file)
     else:
-        final_file = remove_dups_and_excl(merge_file, exclusion_regexes)
+        shutil.copy(temp_file.name, final_file.name)
 
     number_of_rules = settings["numberofrules"]
     output_subfolder = settings["outputsubfolder"]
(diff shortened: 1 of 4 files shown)


updateHostsFile.py
```diff
@@ -259,6 +259,10 @@ def main():
     source_data_filename = settings["sourcedatafilename"]
     no_unified_hosts = settings["nounifiedhosts"]
 
+    settings["targetip"] = (
+        None if settings["targetip"] == "None" else settings["targetip"]
```
ldlac, line 263: I preferred to go with an explicit `None` instead of `""`

StevenBlack:
Hey @ldlac thank you for this!  Looks cool!  Looks thisclose 🤏🏻 to mergable, right out of the box, which is awesome 👍🏻 

Can I please ask for two additional things?

- [ ] Can you please make `-- ip None` case insensitive?  So `-- ip none` will work too.  I *might* go as far as suggesting anything with non-numeric first character following `--ip` should probably work too, but I leave this to you to decide.

- [ ] Can you please add documentation in the [Command line options](https://github.com/StevenBlack/hosts/blob/master/readme_template.md#command-line-options) section of `readme_template.md` file?

Good work!

ldlac:
Good recommendation for the `--ip` flag

I added an insightful explanation in the readme, let me know what you think

ldlac:
By doing so, we'll break ipv6 no?

StevenBlack:
@ldlac good point.

ldlac:
I changed it to `str(settings["targetip"]).lower() == "none"` which will make it case insensitive

StevenBlack:
@ldlac thank you very much!  All tests pass, and all this behaves exactly as expected.

Except for one thing 😄 

With the `--ip none` flag here are the first 50-lines of what is now not a hosts file (see below ⬇️⬇️) 

**Observe:**  we still get that header section with `0.0.0.0 localhost` etc.

**Expected:** (well, I would expect...) no such header, since this is not a hosts file now.  So maybe this PR should also bracket the code that writes the header section.

I also now wonder... should the `--ip none` flag be a super flag, override every other flag, and produce a file named `domains.txt` imstead of `hosts`, and additionally strip all comments and blank lines, and sort the domains too?

I'm willing to be convinced otherwise in any and all these things.

Welcome to Open Source 🙃 

The first 50 lines of the (misnamed) `hosts` file produced by the `--ip none` flag:

```
# Title: StevenBlack/hosts
#
# This hosts file is a merged collection of hosts from reputable sources,
# with a dash of crowd sourcing via GitHub
#
# Date: 19 October 2024 23:10:22 (UTC)
# Number of unique domains: 117,213
#
# Fetch the latest version of this file: https://raw.githubusercontent.com/StevenBlack/hosts/master/hosts
# Project home page: https://github.com/StevenBlack/hosts
# Project releases: https://github.com/StevenBlack/hosts/releases
#
# ===============================================================

127.0.0.1 localhost
127.0.0.1 localhost.localdomain
127.0.0.1 local
255.255.255.255 broadcasthost
::1 localhost
::1 ip6-localhost
::1 ip6-loopback
fe80::1%lo0 localhost
ff00::0 ip6-localnet
ff00::0 ip6-mcastprefix
ff02::1 ip6-allnodes
ff02::2 ip6-allrouters
ff02::3 ip6-allhosts
0.0.0.0 0.0.0.0

# Custom host records are listed here.


# End of custom host records.
# Start StevenBlack

#=====================================
# Title: Hosts contributed by Steven Black
# http://stevenblack.com

ad-assets.futurecdn.net
ck.getcookiestxt.com
eu1.clevertap-prod.com
wizhumpgyros.com
coccyxwickimp.com
webmail-who-int.000webhostapp.com
010sec.com
01mspmd5yalky8.com
0byv9mgbn0.com
ns6.0pendns.org
dns.0pengl.com
```

StevenBlack:
@ldlac an idea: we already have internal support for some of what I suggest above ⬆️ maybe the `--ip none` flag would set these things automatically?

```
--auto, -a            Run without prompting.

--keepdomaincomments, -k
                        Do not keep domain line comments.

--minimise, -m        Minimise the hosts file ignoring non-necessary lines
                        (empty lines and comments).

--noupdate, -n        Don't update from host data sources.

--skipstatichosts, -s
                        Skip static localhost entries in the final hosts file.

--skipstatichosts, -s
                        Skip static localhost entries in the final hosts file.

```

updateHostsFile.py
```diff
@@ -287,19 +302,22 @@ def main():
     merge_file = create_initial_file(
         nounifiedhosts=no_unified_hosts,
     )
-    remove_old_hosts_file(settings["outputpath"], "hosts", settings["backup"])
```
ldlac, line 290: I made few changes here since I introduced options["outputfilename"]

updateHostsFile.py
```diff
@@ -943,14 +966,6 @@ def remove_dups_and_excl(merge_file, exclusion_regexes, output_file=None):
                 if line and not line.startswith("#"):
                     settings["exclusions"].append(line)
 
-    if not os.path.exists(settings["outputpath"]):
```
ldlac, line 946: I changed output_file to required, and I made a few changes here, the file is already ensure created by `remove_old_hosts_file`

updateHostsFile.py
```diff
@@ -906,16 +928,18 @@ def minimise_file(input_file, target_ip, output_file):
     for line in input_file.readlines():
         line = line.decode("UTF-8")
 
-        if line.startswith(target_ip):
-            lines.append(line[: line.find("#")].strip() + "\n")
+        if target_ip is None or line.startswith(target_ip):
```
ldlac, line 931: I don't think it was desired, there was some empty `\n` with the `minimised` flag, this will fix those

ldlac:
Thanks for the feedback, I added 3 comments to support my changes to support the `domains.txt` format

Let me know if you think we should modify the header, I think it's fine even it if it mentions `This hosts file is a merged collection ...`

There's a few things that still bother me, the blacklist, the custom host file, those could still be pass as arg with the `--ip none`. If they contain the default content - only comments - these lines will be stripped

PR #2783, opened 2024-12-15
Add hosts for Draw Joust

marbens-arch:
Add multiple hosts for ads and trackers used by the Draw Joust game
by VOODOO.

### DIFF
--- a/data/StevenBlack/hosts
+++ b/data/StevenBlack/hosts
@@ -2849,3 +2849,31 @@
 
 # Added December 4, 2024
 0.0.0.0 html-load.com
+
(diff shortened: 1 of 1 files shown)


StevenBlack:
Thank you for this Mackus @marbens-arch!

37 category-specific domains with zero overlaps. What's not to love ❤️ about that?

Merging!

marbens-arch:
> Thank you for this Mackus @marbens-arch!
> 
> 37 category-specific domains with zero overlaps. What's not to love ❤️ about that?

I think it's actually 26.

PR #2812, opened 2025-01-30
docs: add more depth to Flake input

toastal:
At present, if a user copy-pastes the current documentation, a version of Nixpkgs will be added to their `flake.lock` lockfile. This is probably not what the user wants as Nixpkgs, the tarball, is quite large & the user likely expects the StevenBlack Flake to be used as a module to just be using the Nixpkgs of the system.

This change should also help new Nix users know how to pin a Nix version if there ever comes a time where the Nix API changes—even if `lib` tries to be stable & backwards-compatible.

Additionally, provide a comment about where & what the input can be—such as a mirror or a local fork for development.

---

Also has been open @ https://gitlab.com/StevenBlack/hosts/-/merge_requests/1

### DIFF
--- a/readme_template.md
+++ b/readme_template.md
@@ -386,7 +386,14 @@ like this:
 
 ```nix
 {
-  inputs.hosts.url = "github:StevenBlack/hosts";
+  inputs = {
+    nixpkgs.url = "github:NixOS/nixpkgs?ref=$YOUR-REF";
+    hosts = {
+      url = "github:StevenBlack/hosts"; # or a fork/mirror
+      inputs.nixpkgs.follows = "nixpkgs";
+    };
+  };
+
   outputs = { self, nixpkgs, hosts }: {
     nixosConfigurations.my-hostname = {
       system = "<architecture>";


StevenBlack:
Thank you for this @toastal.

This repository has 31 different variants, so we have 31 separate `readme.md`, files and they are all auto-generated from a template.

Changes to `readme.md` should be made in [`readme_template.md`](https://github.com/StevenBlack/hosts/blob/master/readme_template.md).

benjamb (approved):

readme_template.md
```diff
@@ -386,7 +386,14 @@ like this:
 
 ```nix
 {
-  inputs.hosts.url = "github:StevenBlack/hosts";
+  inputs = {
+    nixpkgs.url = "github:NixOS/nixpkgs?ref=$YOUR-REF";
```
benjamb, line 390: This could be a little more consistent with the placeholder `<architecture>` below, but it's still an improvement over the existing README.

Pandapip1: `pkgs` is used for the devshell, so you can't _just_ use `lib` here. +1 to this PR as-is.

StevenBlack:
Thank you Gavin @Pandapip1, @toastal, and Ben @benjamb

I'm going to merge this now. It will be live in the next release.

PR #2813, opened 2025-01-30
nix flake refactor: avoid `with lib;` & use `lib.optional*`

toastal:
Upstream Nixpkgs has been pushing the style of removing `with lib;` with the reasoning of clarity around where variables in scope come that hurt readability & static analysis[^1].

While on the topic of `lib`, introduce `lib.optional` & `lib.optionalString` in a few more places to tidy up the code while using common library patterns.

[^1]: https://nix.dev/guides/best-practices#with-scopes

---

Also has been open @ https://gitlab.com/StevenBlack/hosts/-/merge_requests/2

PR #2814, opened 2025-01-30
nix flake: use `lib.mkAfter` put the host list at the end

toastal:
At present, if a user uses `networking.stevenBlockHosts.enable = true` in tandem with `networking.extraHosts`, the `extraHosts` will put be put after which makes it very difficult to see the custom adds with even the `$PAGER` operating a bit slow due to file size. I would propose putting this project’s hosts at the end of the hosts file. Meaning:

```nix
{
  networking = {
    stevenBlockHosts.enable = true;
    extraHosts = ''
      127.0.0.1 myproject.localhost
    '';
  };
}
```

will now output

```
127.0.0.1 localhost
::1 localhost

127.0.0.1 myproject.localhost

# Title: StevenBlack/hosts with the fakenews extension
#
# …
```

NOTE: using `lib.*` like #2813  to avoid a potential merge conflict

---

Also has been open @ https://gitlab.com/StevenBlack/hosts/-/merge_requests/3

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -38,7 +38,7 @@
                 orig = builtins.readFile ("${self}/" + (lib.optionalString (alternatesList != []) alternatesPath) + "hosts");
                 ipv6 = builtins.replaceStrings [ "0.0.0.0" ] [ "::" ] orig;
               in
-              orig + (lib.optionalString cfg.enableIPv6 ("\n" + ipv6));
+              lib.mkAfter (orig + (lib.optionalString cfg.enableIPv6 ("\n" + ipv6)));
           };
         };
 


Pandapip1 (approved):
Not too familiar with the `mkOrder`-like nix functions, but this looks sane to me. Have not tested this, however, and this does seem like it has the potential of breaking things.

toastal:
@Pandapip1 `mkOrder` will let you ‘unprioritize’ an appended string/list. In this case, since this hosts lists is 10s of 1000s of lines long, it is a pain to find your custom settings as even the `$PAGER` will chug on opening a file this long. But also Unix’s `head` is faster than `tail` due to how files work so this could really help those reading their settings.

benjamb:
> Not too familiar with the `mkOrder`-like nix functions, but this looks sane to me. Have not tested this, however, and this does seem like it has the potential of breaking things.

Agreed, changing the order has the potential to change the behaviour for users. Perhaps this could be controlled via another config option?

toastal:
> Agreed, changing the order has the potential to change the behaviour
> for users. Perhaps this could be controlled via another config option?

This is true, but has 2 issue. The 1st being as mentioned, the user has
a hard time getting to the hosts values they have set, but 2nd the first 
match in the hosts files is what the system will use… meaning that if 
the user wants to override xxx.com, they will not be able to since by
default, Nix will append custom values to end. Having used the Python
scripts in the past, the behavior was to append to the StevenBlack hosts 
to the file, not prepend. I think this is a pretty important thing to 
consider & having thought about it, the current behavior could actually
be considered a ‘bug’.

Pandapip1:
I'm gonna assume that the author of this PR has tested this. Assuming it works correctly, this is good to merge and seems like a reasonable change.

toastal:
@Pandapip1 I have tested for several months since I first opened the merge request on GitLab after the maker said GitLab was a supported mirror. …Hence many rebases.

Pandapip1:
Sounds like this should be merged then :shipit:

benjamb:
I suppose given it's the first match that will be used, thus any overrides would need to be at the top, this does make sense. I think any chance of this changing the behaviour for any users is pretty slim.

benjamb (approved):

Pandapip1:
Bump!

Wittionary:
Would love to see this change merged as well :)

30350n:
LGTM! @StevenBlack ([Same comment I wrote on the other PR applies](https://github.com/StevenBlack/hosts/pull/2813#issuecomment-3953293100))

StevenBlack:
Thanks everybody @toastal Bobbe @30350n Witt @Wittionary Ben @benjamb and Gavin @Pandapip1.

Merging.

StevenBlack:
Whoops sorry @toastal can you address the merge conflict here before I merge this?  Sorry about that.

toastal:
ping @StevenBlack

toastal:
If this is merged, I could be freed up to do the minimal changes required to make the Nix code _not_ require experimental features to function.

StevenBlack:
Thanks @toastal.

Merging.

PR #2839, opened 2025-03-14
added 0.0.0.0 rahaporn.com

zerosays:


StevenBlack:
Thank you for this Cyrus @zerosays!

Unfortunately the file you changed is generated from data, and would be clobbered the next time we release amalgamated hosts files from data.

You may want to submit this PR to one of our porn curators, like @sinfonietta for example.

Closing.

PR #2918, opened 2025-07-04
Update release-it to v19.0.3

XhmikosR:
Should be the same but requires Node.js >= 20 since Node.js 18 is EOL.


StevenBlack:
Thanks for this @XhmikosR.  Merging!

PR #2941, opened 2025-08-02
stats.sh: process files in memory instead of on disk

miallo:
Writing the file to the filesystem for all tags is unnecessary if the content can just be parsed in memory. This improves the performance:

Here the output of `time ./stats.sh`:

old
./stats.sh  5.96s user 26.53s system 94% cpu 34.533 total

new
./stats.sh  5.85s user 22.86s system 102% cpu 28.040 total

So a reduction of 6.5s (~20%) is achieved even on a computer with a decent SSD.

### DIFF
--- a/stats.sh
+++ b/stats.sh
@@ -1,12 +1,13 @@
 #!/usr/bin/env bash
 
-echo \n "" > stats.out
+# clear file
+true > stats.out
 
 for TAG_DATE in $(git tag --sort=creatordate  --format='%(refname:short),%(creatordate:short)'); do
   # echo "$TAG_DATE"
   split=(${TAG_DATE//,/ })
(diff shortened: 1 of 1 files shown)


StevenBlack:
Michael @miallo! This is so good!

IMG_3881

StevenBlack:
Now can you do [sourcestats.sh](https://github.com/StevenBlack/hosts/blob/master/sourcestats.sh)? 🙂

Merging.

miallo:
> Now can you do [sourcestats.sh](https://github.com/StevenBlack/hosts/blob/master/sourcestats.sh)? 🙂

Sure thing! (Though today is quite full, so might only be able to do this tomorrow)

PR #2942, opened 2025-08-03
sourcestats: do not checkout files

miallo:
Got a minute to look at this: https://github.com/StevenBlack/hosts/pull/2941#issuecomment-3146893782

But didn't have the time to install `rh` and test it (so please do that before merging) ^^


sourcestats.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
```
miallo, line 1: Needed to change the shell, because process substitution (`<()`) isn't POSIX compliant… Alternatively I'd need to figure out a way do this differently with `rh`. If `bash` (or `zsh`) isn't an option, could you give me a hint what `rh` is and maybe even a link to a man page? I only found https://github.com/twigly/rust-http-cli

StevenBlack:
Thank you for this Michael @miallo!  This is just great! ❤️

Merging!

infamous28:
How does the shoppafiy ai work

PR #2947, opened 2025-08-08
added domains from playstream.media

cgbarros:
Got some video ads and, upon inspecting my pi hole logs, I saw they were coming from playstream.media. Their website says they "built a connection between content creators marketeers and the audience" (emphasis on marketeers). So, I added all the domains and subdomains it pulled


StevenBlack:
Obrigado Caio @cgbarros!

Merging.

PR #2950, opened 2025-08-17
Add new ai-chat list to block unsafe AI chatbot platforms

OmerArfan:
Introduce hosts/alternates/ai-chat, a new list targeting unsafe AI chatbot platforms.  
Includes domains hosting explicit/NSFW chatbots lacking safeguards for minors.


StevenBlack:
Thank you for this Omer @OmerArfan. I appreciate the effort you've put into this.

I'm going to decline this pull request for the following reasons.

Firstly, we rely on curators upstream who have a track record of maintaining their lists. We are a derivative work from the lists they maintain.

Secondly, everything in the `alternates` subfolder is generated from subdolders of the `extensions` folder.  In other words, if we were to include an AI or chatbot list, that's where it would go, and it would point to an external curator of AI and/or chatbot domains.

Thirdly it feels too soon to have an unsafe AI chatbot category. If we were to have such a concern, why would it not fit squarely with the others within the broader adware/malware that's our base amalgamated list?

Therefore at this juncture I decline this PR. That said, thank you for submitting this.

OmerArfan:
Hi Steven @StevenBlack , thank you for taking the time to review this PR and for explaining your view. Even if it was declined, I really appreciate that you acknowledged it.

My main intention here was to raise awareness — as a teenager myself, I’ve seen unsafe AI chatbots quietly becoming a problem among younger users. UNICEF has also highlighted this issue, noting that while AI chatbots are marketed as emotional support, they are poorly regulated and in some cases even engage in explicit conversations: https://www.unicef.org/innocenti/stories/risky-new-world-techs-friendliest-bots

Thanks again for your work on this project!

PR #2960, opened 2025-09-17
CI: pin actions

XhmikosR:
This is considered a better practice security-wise since hashes are immutable. Dependabot handles updating hashes the same way, so there should be no problem updating actions.

We could also group dependabot updates so that you get one PR with all updates @StevenBlack. Let me know if you prefer that.


StevenBlack:
@XhmikosR that's interesting!  I didn't know about hash-binding versions.

Merging!

PR #2978, opened 2025-10-05
Add CZ fakenews sites

vsedmik:
A couple of Czech fakenews sites


StevenBlack:
Thank you for this Vladimir @vsedmik.

I must decline this pull request because the domains have been added to `data/StevenBlack/hosts` which gets included in most of our hosts lists, fake-news of not.

The fake news list is curated by @marktron and gets included only in fake news variants of our lists.  

@marktron are you able to evaluate these domains for inclusion in your list?

Closing.

PR #2980, opened 2025-10-09
Dependabot: group action updates

XhmikosR:
Now you will get on PR with all action updates @StevenBlack.

BTW you should merge  the currently opened dependabot PRs, they are OK.


StevenBlack:
Thank you as always @XhmikosR.

Merging!

PR #3013, opened 2025-11-22
Dockerfile: fix LegacyKeyValueFormat warnings

XhmikosR:
Should fix these:

```
Legacy key/value format with whitespace separator should not be used: Dockerfile#L3
LegacyKeyValueFormat: "ENV key=value" should be used instead of legacy "ENV key value" format More info: https://docs.docker.com/go/dockerfile/rule/legacy-key-value-format/

Legacy key/value format with whitespace separator should not be used: Dockerfile#L11
LegacyKeyValueFormat: "ENV key=value" should be used instead of legacy "ENV key value" format More info: https://docs.docker.com/go/dockerfile/rule/legacy-key-value-format/
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 FROM docker.io/python:3-alpine
 
-ENV IN_CONTAINER 1
+ENV IN_CONTAINER=1
 
 RUN apk add --no-cache git sudo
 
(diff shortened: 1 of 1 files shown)


StevenBlack:
Thank you for this @XhmikosR.  Good catch.

Merging.

PR #3016, opened 2025-11-23
Remove donmai.us

Parsa307:
Is an innocent domain. default backend - 404

Image

### DIFF
--- a/extensions/porn/sinfonietta/hosts
+++ b/extensions/porn/sinfonietta/hosts
@@ -12249,7 +12249,6 @@
 0.0.0.0 donfreeporn.com
 0.0.0.0 donkparty.com
 0.0.0.0 donlot.xyz
-0.0.0.0 donmai.us
 0.0.0.0 betabooru.donmai.us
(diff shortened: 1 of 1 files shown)


waifucon:
Oops, I did also made an change to alternatives directory. 😅

Hope it won't break anything.

Force pushed. to undo those changes.

StevenBlack:
Thank you for this Parsa @Parsa307.

Similar to my comment as #3017:

The file you modified come from upstream, specifically from one of our porn list curators.

So any changes here will be overwritten the next time we refresh from upstream.

Moreover this domain is only included in our porn-related block lists.

Therefore I decline this pull request.

Closing.

Parsa307:
@Sinfonietta, could you take a look at this PR:https://github.com/Sinfonietta/hostfiles/pull/359

i've seen you've responded to some mentions at this repo, that domain as of currently does respond with default backend - 404 however if the user is using ublock origin with the host file it does block safebooru.donmai.us (SFW).

PR #3017, opened 2025-11-23
Remove cdn.donmai.us

Parsa307:
Blocking cdn.donmai.us affects safebooru.donmai.us The images won't show up.


waifucon:
Oops, I did also made an change to alternatives directory 😅

I forced pushed and made the modifications to extensions directory only.

StevenBlack:
Thank you for this @Parsa307 

The files you modified come from upstream, specifically from two different porn list curators, who consider that the domain `cdn.donmai.us` vehicles porn.

So any changes here will be overwritten the next time we refresh from upstream.

Moreover this domain is only included in our porn-related block lists.

Therefore I decline this pull request.

Closing.

PR #3040, opened 2025-12-26
Added over 1400 gambling domains

LeonidNougat:
Never commited anything in my life but this Christmas I locked in. Hopefully this works and will add to scratching the surface of what's currently an epidemic.

PS: Everything has been (re)sorted alphabetically.


StevenBlack:
Thank you very much for this Leonid @LeonidNougat.

Unfortunately the file you pushed doesn't originate here.  It belongs to the gambling list maintained by @sinfonietta.  For each release we pull the latest lists maintained by several external curators like @sinfonietta, and combine them in various ways to produce a variety of specialty lists.

So if you could please submit this to @sinfonietta here: https://github.com/Sinfonietta/hostfiles  

Also, when you submit a PR, it's a good idea to organize it so review is easy.  Making 1400 additions and gratuitously sorting the results creates an incredible amount of work for reviewers since changes are scattered in a thousand different locations.

Closing.

PR #3042, opened 2025-12-27
Remove basecamp-shop.com.

jorgii:
It is a Bulgarian outdoor equipment store- https://basecamp-shop.com.

I cannot image that it serves ads.

### DIFF
--- a/hosts
+++ b/hosts
@@ -22303,7 +22303,6 @@ ff02::3 ip6-allhosts
 0.0.0.0 bajkowewnetrzakids.pl
 0.0.0.0 banvar.com
 0.0.0.0 bartekzukiewicz.pl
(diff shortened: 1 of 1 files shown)


StevenBlack:
Thank you for this George @jorgii.

Unfortunately the domain `basecamp-shop.com` comes to us from upstream, and the file changed in this PR gets overwritten each time we create a new release.

You can make your case for domain `basecamp-shop.com` with the KADhosts list maintainers here: https://github.com//FiltersHeroes/KADhosts/issues

Closing.

jorgii:
I asked about this in the upstream and it was removed: 
https://github.com/FiltersHeroes/KAD/issues/2699

PR #3079, opened 2026-02-25
Add missing lib to Nix flake

Hasnep:
This was accidentally removed in a recent change and broke the flake.


StevenBlack:
Thank you very much for this Hannes @Hasnep.

I'm just going to double check with those involved.

Bobbe @30350n and @toastal any thoughts on this?

30350n:
LGTM! This should be merged, without out it the current flake is completely broken. (Seems like the line in question came from a PR merged after #2813 s creation, which is why this slipped through. We should have checked #2813 for correctness before merging again.)

StevenBlack:
Thank you Bobbe @30350n and Hannes @Hasnep.

Merging!

PR #3157, opened 2026-05-31
removed macos host lines

yuannan:
This is for NixOS and therefore uneeded and causes issues with dnsmasq.

This is a simple modification that removes these OSX specific lines from the host files.

This is required on NixOS as it is immutable, therefore the file needs to be modified in place before it can be installed.

I've checked the file and there seems to no extra collateral damage. All comments start with # and therefore cannot contain `%`, and even if they did it's a harmless removal.

The rest of the lines are actual host file information which we want to autoremove since it's incompatible with NixOS.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -41,9 +41,19 @@
                 orig = builtins.readFile (
                   "${self}/" + (lib.optionalString (alternatesList != [ ]) alternatesPath) + "hosts"
                 );
-                ipv6 = builtins.replaceStrings [ "0.0.0.0" ] [ "::" ] orig;
+                filterHosts = text:
+                  builtins.concatStringsSep "\n"
+                    (builtins.filter
+                      (line:
+                        let first = builtins.elemAt (lib.splitString " " line) 0;
+                        in !(lib.hasInfix "%" first)
+                      )
+                      (lib.splitString "\n" text)
+                    );
+                filtered = filterHosts orig;
+                ipv6 = builtins.replaceStrings [ "0.0.0.0" ] [ "::" ] filtered;
               in
-              lib.mkAfter (orig + (lib.optionalString cfg.enableIPv6 ("\n" + ipv6)));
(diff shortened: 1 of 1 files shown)


StevenBlack:
Thanks for this Yuannan @yuannan.

Any comment Bobbe @30350n?

yuannan:
@StevenBlack I've also noticed that it will crash dnsmasq if some of the lines are over 57 chars total.

You can check out my `short` branch.

I presume that is too much of a modification for you to merge?

It only takes off a few hundred hosts from the ~100k in my testing.

I've sent a report into dnsmasq but I'm not too hopefully that it'll be fixed anytime soon due to a bombardment of slop reports.

The workaround for me was to just add `no-hosts` to my libvirtd default network configs in the meantime.

StevenBlack:
Thanks for the clarification Yuannan @yuannan.

I just want opinions from some NixOS users I know.

StevenBlack:
Ping again Bobbe! @30350n 

Any comment on this?

30350n:
What does this remove exactly / what does `%` do and how is it related/specific to macos and dnsmasq?  I'm not that well acquainted with the exact workings of the `hosts` file.

The nix code generally looks fine from a first glance.

StevenBlack:
Thanks again for this Yuannan @yuannan.

I've decided to decline this pull request because I see no evidence that this PR addresses "OSX specific lines" at all.

As far as I can tell, the default MacOS hosts file is:

```
##
# Host Database
#
# localhost is used to configure the loopback interface
# when the system is booting. Do not change this entry.
##
127.0.0.1       localhost
255.255.255.255 broadcasthost
::1             localhost
```

Closing.

yuannan:
@StevenBlack This is because the hostlines by default have this:

```
fe80::1%lo0 localhost
```

`%lo0` is only used on MacOS as far as I'm aware, Linux uses just `lo`.

The extra `%` will trip up some systems.

PR #3189, opened 2026-07-05
sourcestats.sh: check for required tools

XhmikosR:
For consistency with stats.sh and because I personally don't have `rh` available and it wipes a few files needlessly :)


StevenBlack:
Yeah @XhmikosR that's a good catch 🙂

I should document that better.

The `rh` piece is the compiled binary of [rhosts](https://github.com/StevenBlack/rhosts) which is a personal side-project, currently rather dormant, to supplement or replace hosts tooling in Rust.

I'm loving Rust, by the way.

Merging!

PR #3192, opened 2026-07-09
add domains associated with Ad-Shield / html-load.com

data-liberator:
Adds some domains used by Ad-Shield. Note that some sites using Ad-Shield (such as Yahoo News) may choose to break if they cannot contact Ad-Shield and pull down ads.


infamous28 (approved):

StevenBlack:
Thank you for this @data-liberator.

Merging.

PR #3250, opened 2026-09-20
Update dependabot config

XhmikosR:
* group minor and patch updates
* add npm
* switch to monthly schedule for everything


StevenBlack:
Thank you for this @XhmikosR, as always.

Merging.

